Issue #37, opened 2025-08-05
Bug: api publication no-op changes continually detected


jdw6359:
I have noticed that when api publications are defined, updates to those entities are picked up by plans, even after a successful apply.

To repro, you can use the minimalistic config below

```yaml
portals:
  - ref: foobar
    name: "Example Developer Portal"
    display_name: "Example Developer Portal"
    description: "Example Developer Portal"
    authentication_enabled: true
    rbac_enabled: false
    auto_approve_developers: false
    auto_approve_applications: false
    default_api_visibility: "private"
    default_page_visibility: "private"

apis:
  - ref: example-api
    # Load metadata from OpenAPI spec using !file tags
    name: !file apis/example-api/spec.yaml#info.title
    description: !file apis/example-api/spec.yaml#info.description
    version: !file apis/example-api/spec.yaml#info.version
    labels:
      foo: bar
    versions:
      - ref: example-api-v1_0_0
        version: "1.0.0"
        spec: !file apis/example-api/spec.yaml
        
    publications:
      # Public publication for external developers
      - ref: example-api-private-publication
        portal_id: foobar
        visibility: private
        auto_approve_registrations: false
```

after this is cleanly applied, subsequent plans will result in the following:

```
➜  developer-portal git:(main) kongctl plan -f portal.yml --mode apply
{
  "metadata": {
    "version": "1.0",
    "generated_at": "2025-08-05T16:17:53.357012Z",
    "generator": "kongctl/dev",
    "mode": "apply"
  },
  "changes": [
    {
      "id": "1:u:api_publication:example-api-to-foobar",
      "resource_type": "api_publication",
      "resource_ref": "example-api-to-foobar",
      "resource_id": "<>:<>",
      "action": "UPDATE",
      "fields": {
        "auth_strategy_ids": null,
        "portal_id": "<>"
      },
      "references": {
        "api_id": {
          "ref": "example-api",
          "id": "<>,
          "lookup_fields": {
            "name": "Tasks API"
          }
        },
        "portal_id": {
          "ref": "foobar",
          "id": "<>",
          "lookup_fields": {
            "name": "Example Developer Portal"
          }
        }
      },
      "parent": {
        "ref": "example-api",
        "id": "<>"
      },
      "namespace": "default"
    }
  ],
  "execution_order": [
    "1:u:api_publication:example-api-to-foobar"
  ],
  "summary": {
    "total_changes": 1,
    "by_action": {
      "UPDATE": 1
    },
[...]

PR #40, opened 2025-08-05
fix: resolve api document slug generation issue (#36)

### LINKED ISSUE #36: `kongctl apply` attempting to create api_document that already exists
When running `kongctl apply -f <>` I get the below output:

Note: It is detecting an `api_document: main` for creation, but the `api_document` already exists. As a result, the creation will 409 with a resource conflict

```
SUMMARY
----------------------------------------------------------------------
  Total changes: 3
  Namespaces affected: 1
  Resources to create: 1
  Resources to update: 2

  Resource breakdown:
    api_document: 1
    api_publication: 1
    portal_page: 1

CONFIRM?
----------------------------------------------------------------------
Do you want to continue? Type 'yes' to confirm: yes

Applying changes:
[1/3] [namespace: default] Updating portal_page: home... ✓
[2/3] [namespace: default] Creating api_document: main... ✗ Error: API error during create of api_document '[unknown]': failed to create API document: {"status":409,"title":"Resource Conflict","type":"https://kongapi.info/konnect/resource-conflict","instance":"kong:trace:4259404297542154812"}
[3/3] [namespace: default] Updating api_publication: example-api-to-seatgeek... ✓

Complete.
Applied 2 changes.

Errors:
  • api_document [unknown]: API error during create of api_document '[unknown]': failed to create API document: {"status":409,"title":"Resource Conflict","type":"https://kongapi.info/konnect/resource-conflict","instance":"kong:trace:4259404297542154812"}
```

rspurgeon:
## Summary

This PR fixes issue #36 where API documents without explicit slugs were incorrectly planned as CREATE instead of UPDATE operations.

### Problem
When API documents didn't have an explicit `slug` field, the planner couldn't match them with existing resources because:
- The Konnect API automatically generates slugs from titles using `slugify(title)`
- kongctl wasn't generating matching slugs, causing the planner to fail resource matching
- This resulted in UPDATE operations being incorrectly planned as CREATE operations

### Solution
- Implemented automatic slug generation in `SetDefaults()` for API documents when slug is not provided
- Created `util.GenerateSlug()` that exactly matches Konnect server's slugify behavior:
  - Unicode normalization (NFKD) for characters like é→e, ñ→n
  - CamelCase word splitting (XMLParser → xml-parser)
  - Special character removal and hyphenation
- Added comprehensive tests including edge cases from the server implementation

### Additional Improvements
While investigating this issue, I also implemented `GetByID` methods for related child resources to improve resource validation:
- API Documents: Uses parent API context for lookups
- Portal Snippets: Uses parent portal context for lookups  
- Portal Custom Domains: Handles singleton resource pattern

### Testing
- Verified API documents without explicit slugs now properly match existing resources
- Tested with various title formats including Unicode characters and camelCase
- All quality checks pass (build, lint, test)
- Manually tested with the example files that reproduced the issue

Fixes #36


rspurgeon:
@jdw6359 looking at this to resolve #36, but I believe it may resolve #37 as well. We can build a release to validate with your use case

PR #59, opened 2025-08-27
fix: enable cross-namespace portal references in API publications

### LINKED ISSUE #58: Apply failure when API publications reference portal in different namespace
## Summary

Applies of an api publication will fail when the portal we are publishing to exists in a different namespace from the api that is being published

## Details

I have a configuration where we define a portal, and have multiple APIs that get published to that portal via a `publication`

This is defined all in a single file, so it is something that should work without `external` resources.

The API in `namespace: default` (same as the portal) has no issues being published to the portal via `publications[].portal_id`

The second API is in `namespace: foobar` (different than the portal) and is having issues being applied.

The apply fails to create the publication with a ref lookup error:

```
...
Errors:
  • api_publication [unknown]: failed to resolve portal reference: portal not found: ref=foobar, looked up by name=foobar
```

What's interesting is that the plan seems to resolve this cross namespace ref without a problem:
```
    {
      "id": "7:c:api_publication:foobar-private-publication",
      "resource_type": "api_publication",
      "resource_ref": "foobar-private-publication",
      "action": "CREATE",
      "fields": {
        "portal_id": "our-portal",
        "visibility": "private"
      },
      "references": {
        "api_id": {
          "ref": "foobar",
          "lookup_fields": {
            "name": "Foobar Name"
          }
        },
        "portal_id": {
          "ref": "our-portal",
          "lookup_fields": {
            "name": ""
          }
        }
      },
      "parent": {
        "ref": "foobar",
        "id": ""
      },
      "namespace": "platform-foobar",
      "depends_on": [
        "2:c:api:foobar"
      ]
    }
```

## Reproduction

To reproduce, you can define portal, and two APIs in a single file. Make sure to scope one of the APIs to a namespace other than `default`, and observe that the plan is generated fine, but the apply will fail due to a ref resolution error

rspurgeon:
## Summary

Fixes issue #58 where API publications fail when referencing portals in different namespaces.

## Problem

When creating API publications that reference portals in different namespaces, the planner would fail with "portal with ref 'test-portal' not found" even though the portal existed. This occurred because the planner only had access to namespace-filtered resources and couldn't see portals from other namespaces.

## Root Cause

The `planAPIPublicationCreate` function in `api_planner.go` was using namespace-filtered portal resources (`p.desiredPortals`) to look up portal references. When the referenced portal was in a different namespace, the lookup would fail.

## Solution

- Added global lookup methods to `ResourceSet` that search across all namespaces
- Added namespace-filtered access methods to `ResourceSet` for existing functionality  
- Updated planner to use single `ResourceSet` with both global and filtered capabilities
- Fixed portal reference resolution to use global lookup while maintaining namespace isolation for planning operations

## Changes

- **internal/declarative/resources/types.go**: Added global and namespace-filtered resource access methods
- **internal/declarative/planner/api_planner.go**: Fixed portal lookup to use global method
- **internal/declarative/planner/planner.go**: Refactored to use single ResourceSet
- **internal/declarative/planner/base_planner.go**: Updated methods to support namespace filtering
- **internal/declarative/planner/portal_planner.go**: Updated to use public GetNamespace function
- **internal/declarative/planner/auth_strategy_planner.go**: Updated to use public GetNamespace function

## Testing

- All existing tests pass
- Build succeeds with no compilation errors
- Solution is general enough to support cross-namespace references for all resource types

Resolves #58

### DIFF
--- a/internal/declarative/planner/api_planner.go
+++ b/internal/declarative/planner/api_planner.go
@@ -46,7 +46,9 @@ func (a *apiPlannerImpl) PlanChanges(ctx context.Context, plannerCtx *Config, pl
 		return err
 	}
 	
-	if err := a.planner.planAPIPublicationsChanges(ctx, plannerCtx, a.GetDesiredAPIPublications(namespace), plan); err != nil {
+	if err := a.planner.planAPIPublicationsChanges(
+		ctx, plannerCtx, a.GetDesiredAPIPublications(namespace), plan,
+	); err != nil {
 		return err
 	}
 	
--- a/internal/declarative/planner/api_planner.go
+++ b/internal/declarative/planner/api_planner.go
@@ -29,8 +29,11 @@ func NewAPIPlanner(base *BasePlanner) APIPlanner {
 
 // PlanChanges generates changes for API resources and their child resources
 func (a *apiPlannerImpl) PlanChanges(ctx context.Context, plannerCtx *Config, plan *Plan) error {
+	// Get namespace from planner context
+	namespace := plannerCtx.Namespace
+	
 	// Debug logging
-	desiredAPIs := a.GetDesiredAPIs()
+	desiredAPIs := a.GetDesiredAPIs(namespace)
 	a.planner.logger.Debug("apiPlannerImpl.PlanChanges called", "desiredAPIs", len(desiredAPIs))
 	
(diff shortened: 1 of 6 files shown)


jdw6359:
totally - this worked as expected!

Issue #201, opened 2025-11-11
Feat: Support Developer Portal Application resources


rspurgeon:
Applications are a read + delete only resource in Konnect API, Users create and manage resources themselves inside portals.  This task is to implement GET + DELETE support for a `portal_application` resource in `kongctl`.

https://developer.konghq.com/api/konnect/portal-management/v3/#/operations/list-applications

```
kongctl get portal applications
kongctl get portal application --portal-id <portal-id> <application-id>
kongctl get portal application --portal-name <portal-name> <application-id>
kongctl get portal application --portal-name <portal-name> <application-name>
kongctl delete portal application --portal-name <portal-name> <application-id>
```

PR #247, opened 2025-11-23
Feat: Adds Konnect Portal Applications support

rspurgeon:
Includes e2e tests with full developer workflow for developer and app registration


rspurgeon:
Resolved #201

PR #360, opened 2026-02-05
Fix: Makes brittle portal applications test off by default and optional

rspurgeon:


test/e2e/scenarios/portal/applications/scenario.yaml
```diff
@@ -2,6 +2,16 @@ baseInputsPath: ../../../testdata/declarative/portal/applications
 
 log-level: debug
 
+test:
```
harshadixit12: The naming can be a bit more clear IMO - `enabledEnv` -> `enabledByEnvVar`, `requiresEnv` -> `requiredEnvVars`

I'd interpret `enabledEnv` as the environment in which this test is enabled, which does not seem to be the intent here.

PR #411, opened 2026-02-16
refactor: introducing base resource and generic methods to reduce boilerplate code

Prashansa-K:
### Summary
This change reduces boilerplate in declarative resource implementations 
by introducing shared patterns.

Changes
1. `BaseResource` struct for Top-level resources
- Provides common fields: ref, kongctl, konnectID
- Provides 6 methods are inherited: GetRef(), GetKonnectID(), SetKonnectID(), IsExternal(), TryMatchByName(), GetKonnectMonikerFilter()
By default, IsExternal() returns false. Resources that need External block can override this method.
- Applied to 7 top-level resources: API, Portal, ControlPlane, OrganizationTeam, AuthStrategy, EventGatewayControlPlane, CatalogService

2. tryMatchByField()
- Simplifies `TryMatchKonnectResource` for child resources.
- Centralizes reflection-based field matching logic
- Applied to 9 child resources: APIPublication, APIVersion, APIDocument, PortalPage, PortalSnippet, PortalTeam, PortalEmailTemplate, EventGatewayVirtualCluster, EventGatewayBackendCluster

3. tryMatchByNameWithExternal()
- Simplifies `TryMatchKonnectResource` for resources that have External block

Any entity that requires custom logic can override the methods.
BaseResource wasn't applied for child resources as they don't require all
the common fields and have custom structs and methods. It was felt better to keep them as
it is.

Benefits
- Less boilerplate: New resources require fewer copy-pasted methods
On an average, each new top-level resource needs 6 identical functions. (~30 LOC per resource)
- Centralized logic: Reflection/matching code in one place. This makes maintenance easier.
Each resource repeats ~20 lines of reflection code.
- Future: Enables us to explore other golang patterns for further simplification

### Documentation

- [ ] Link to the website [documentation PR](https://github.com/Kong/docs.konghq.com/pull/XXX)

### Testing

- [ ] Unit tests
- [ ] E2E tests

### DIFF
--- a/internal/declarative/resources/control_plane.go
+++ b/internal/declarative/resources/control_plane.go
@@ -128,7 +128,14 @@ func (c *ControlPlaneResource) IsExternal() bool {
 
 // TryMatchKonnectResource attempts to match this resource with a Konnect resource
 func (c *ControlPlaneResource) TryMatchKonnectResource(konnectResource any) bool {
-	return c.TryMatchByName(c.Name, konnectResource, matchOptions{})
+	id, ok := tryMatchByNameWithExternal(c.Name, konnectResource, matchOptions{
+		sdkType: "ControlPlane",
+	}, c.External)
+
+	if ok {
+		c.SetKonnectID(id)
+	}
+	return ok
 }
 
 // IsGroup returns true when the control plane represents a control plane group.


rspurgeon:
@Prashansa-K Thanks for the refactoring work here. I have the following thoughts, please correct me if I'm misunderstanding.

- This change broadens `_external` acceptance via `BaseResource`, but API/catalog/EGW planners do not implement external resolution or skip logic. 
- `Validate()` for API/catalog/EGW should validate _external (or reject it), otherwise _external: {} silently passes. I think we should either add full external lifecycle support or reject `_external` on unsupported resource types at load time. 
- Loader namespace/protected handling should explicitly treat unsupported _external_ resources as an error, not managed resources with defaults.

Prashansa-K:
Thanks for pointing that out Rick. I had missed that.
At the moment, I will decompose the Base struct into two. This would be easier for us without adding extra validation layers on top.
We can track the resources that need external support and are missing separately.

Making more changes in this PR.

internal/declarative/resources/base.go
```diff
@@ -0,0 +1,116 @@
...
+// MatchOptions configures how TryMatchByName performs matching.
+type MatchOptions struct {
+	// SDKType is the embedded struct name for SDK responses (e.g., "Portal", "APIResponseSchema").
+	// Used when the SDK response embeds a type with Name/ID fields.
+	SDKType string
+}
+
+// BaseResourceCore provides the minimal common fields for declarative resources.
+// Use this for resources that don't support external resource references.
+type BaseResourceCore struct {
```
rspurgeon: Subjectively, the names of `BaseResourceCore` and `BaseResource` don't help the reader (or future developer / agent) know the differences between the two. 

IMO `BaseResource` for this struct is sufficient, but we then change slightly the other struct to represent more what it does, which is bring external fields to the struct.  I will comment there as well. 


```suggestion
type BaseResource struct {
```

PR #443, opened 2026-02-19
Add kongctl lint command

mheap:
### Summary

Adds a spectral-compatible linting engine to `kongctl` (the same as `deck file lint`)

```bash
❮ go run main.go lint -f /tmp/aaa/config.yaml -r /tmp/ruleset.yaml
Linting Violations: 1
Failures: 1

/tmp/aaa/config.yaml:8:19: [error] display_name '{{value}}' must start with 'Konnect'
Error: found 1 linting violation(s) at or above "error" severity
exit status 1
```

```bash
❮ go run main.go lint -R -f /tmp/aaa -r /tmp/ruleset.yaml
Linting Violations: 1
Failures: 1

/tmp/aaa/config.yaml:8:19: [error] display_name '{{value}}' must start with 'Konnect'
Error: found 1 linting violation(s) at or above "error" severity
exit status 1
```

### Full changelog

* Implement `kongctl lint` subcommand

### Testing

- [x] Unit tests
- [ ] E2E tests

### DIFF
--- a/internal/lint/files.go
+++ b/internal/lint/files.go
@@ -8,15 +8,6 @@ import (
 	"strings"
 )
 
-// readFile reads a file from disk and returns its bytes.
-func readFile(path string) ([]byte, error) {
-	data, err := os.ReadFile(path)
-	if err != nil {
-		return nil, fmt.Errorf("reading file %q: %w", path, err)
-	}
-	return data, nil
-}
-
 // CollectFiles resolves file paths from the provided sources. If a
 // path is a directory, it collects all YAML/YML files within it.
 // When recursive is true, it walks subdirectories as well.
@@ -30,6 +21,9 @@ func CollectFiles(paths []string, recursive bool) ([]string, error) {
 		}
 
 		if !info.IsDir() {
+			// Files specified directly are included regardless of extension.
+			// This allows linting non-YAML formats (e.g. JSON OpenAPI specs)
+			// when the caller explicitly provides the path.
 			files = append(files, p)
 			continue
 		}
--- a/internal/lint/lint.go
+++ b/internal/lint/lint.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"os"
 	"sort"
 
 	"github.com/daveshanley/vacuum/motor"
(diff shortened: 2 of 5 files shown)


internal/lint/lint.go
```diff
@@ -0,0 +1,326 @@
...
+			Severity: sev,
+			RuleID:   ruleID,
+			Line:     line,
+			Column:   col,
+			File:     fileName,
+		})
+	}
+
+	// Sort results by file, then line, then column for consistent output
+	sort.Slice(results, func(i, j int) bool {
```
rspurgeon: **Design: Redundant sort when called from `Files`**

`Bytes` sorts its results here, but `Files` re-sorts the aggregated results at line 254. When `Bytes` is invoked via `Files`, this per-file sort is wasted work.

Consider removing the sort from `Bytes` (documenting that results are in engine order) and only sorting once at the aggregation boundary in `Files`.

internal/lint/files.go
```diff
@@ -0,0 +1,75 @@
...
+func CollectFiles(paths []string, recursive bool) ([]string, error) {
+	var files []string
+
+	for _, p := range paths {
+		info, err := os.Stat(p)
+		if err != nil {
+			return nil, fmt.Errorf("cannot access %q: %w", p, err)
+		}
+
+		if !info.IsDir() {
```
rspurgeon: **Design: Non-YAML files silently accepted when specified directly**

The YAML extension filter only applies when traversing a *directory*. A file path specified directly is always included regardless of extension (e.g. `kongctl lint -f config.json` will attempt to lint the JSON file).

This may be intentional (vacuum supports JSON OpenAPI specs), but it is undocumented and inconsistent with the YAML-only filtering applied to directory entries. Consider adding a comment, or applying the same filter uniformly.

rspurgeon (commented):
Code review for the  command implementation. Overall the code is clean, well-tested, and follows project conventions. A few issues below ranging from a clear bug to design and test concerns.

internal/lint/lint.go
```diff
@@ -0,0 +1,326 @@
...
+		if results[i].File != results[j].File {
+			return results[i].File < results[j].File
+		}
+		if results[i].Line != results[j].Line {
+			return results[i].Line < results[j].Line
+		}
+		return results[i].Column < results[j].Column
+	})
+
+	return &Output{
```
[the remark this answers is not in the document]
rspurgeon: Nil Results slice (lint.go)
  - Initialized results = []Result{} instead of var results []Result in Bytes() so Output.Results marshals as [] not null in JSON when there are no violations

PR #566, opened 2026-03-13
feat: team members

Prashansa-K:
This PR adds support for member additions in a team.
We can add a pre-existing user or a system-account to a team.

This is what the declarative UX looks like:
```
organization:
  teams:
    - ref: backend-team
      name: "Backend Team"

      members: # a grouping concept for better readability and collapsing yaml needs
        users:
         - ref: alice
           name: Alice # we can also support id, email for lookups  

        system_accounts:
         - ref: bot-account
           name: Bot # we can also support id for lookup
```

Get and list commands follow the same heirarchy.

The following commands would output all members of a team - users and system-accounts
- kongctl get org teams members --team-name teamA
- kongctl get org teams members --team-id <some-uuid>

We can classify them more into specific members and filter based on names, ids and emails (in case of users).
- kongctl get org teams members users --team-name teamA alice
- kongctl get org teams members users --team-name teamA alice@gmail.com
- kongctl get org teams members system-account --team-name teamA bot

For tests, I have used system-accounts in the scenarios as we can't readily create and delete users. 
So, I am testing all functionalities using system-accounts.

Support for adopt and dump would be added in a separate PR.

### DIFF
--- a/internal/cmd/root/products/konnect/declarative/declarative.go
+++ b/internal/cmd/root/products/konnect/declarative/declarative.go
@@ -2279,5 +2279,6 @@ func createStateClient(kkClient helpers.SDKAPI) *state.Client {
 
 		// Organization APIs
 		OrganizationTeamAPI: kkClient.GetOrganizationTeamAPI(),
+		TeamMembershipAPI:   kkClient.GetTeamMembershipAPI(),
 	})
 }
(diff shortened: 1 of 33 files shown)


rspurgeon:
@Prashansa-K closing this in favor of other work being done to close up the Konnect Organization support. I changed the shape of the input declarative configuration slightly.

Issue #216, opened 2025-11-12
Feat: Event Gateway Virtual Cluster Produce Policies resource support


rspurgeon:
https://developer.konghq.com/api/konnect/event-gateway/v1/#/operations/list-event-gateway-virtual-cluster-produce-policies

Issue #214, opened 2025-11-12
Feat: Event Gateway Virtual Cluster Policies resource support


rspurgeon:
https://developer.konghq.com/api/konnect/event-gateway/v1/#/operations/list-event-gateway-virtual-cluster-cluster-level-policies

PR #702, opened 2026-03-30
Feat: Event Gateway produce policy

harshadixit12:
## Summary
In this PR, we are adding support for Event Gateway Virtual Cluster Produce Policy - 
1. Declarative Management
2. Imperative Get
3. Dump
4. View

Note:
Resource key is `produce_policies` when nested under `virtual_clusters`, and `event_gateway_virtual_cluster_produce_policies` when declared at root level.

Example for imperative Get:
```
kongctl get egw vc produce-policies --gateway-id $EGW_ID --virtual-cluster-id $VC_ID --pat $PAT

kongctl get egw vc produce-policies $POLICY_ID --gateway-id $EGW_ID --virtual-cluster-id $VC_ID --pat $PAT
```

## Full changelog
[Implement https://github.com/Kong/kongctl/issues/214](https://github.com/Kong/kongctl/issues/216)

## Testing
- [x] Unit tests
- [x] E2E tests

### DIFF
--- a/test/e2e/scenarios/event-gateway/dump/scenario.yaml
+++ b/test/e2e/scenarios/event-gateway/dump/scenario.yaml
@@ -15,6 +15,7 @@ vars:
   egwName: egw-cp-for-dump-test
   bcRef: default-backend-cluster
   vcRef: default-virtual-cluster
+  producePolicyName: default-produce-policy-name-for-dump-test
 
 defaults:
   mask:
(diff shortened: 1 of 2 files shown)


rspurgeon:
> @rspurgeon We don't at the moment, it was on my list once the remaining resources were added. However, as a rule of thumb - for every PR, I do apply -> dump -> apply again to verify that dump works, and doesn't produce drift.

@harshadixit12 Please add to the e2e test flow for these two consume/produce policies a dump && adopt && apply scenario and then we can push these through.  Thank you

PR #908, opened 2026-04-27
Security: added url validation for login and auth flows

Prashansa-K:
For: https://konghq.atlassian.net/browse/APIOP-7

This was marked by Aikido as a critical vulnerability in a sub-issue.
On inspection, it doesn't look like a critical issue. 
Chances of a SSRF attack seem to be low
as the input is client controlled.
Erring on the side of caution, this change adds a validation layer
on top for the URLs. This enables an early exit in case of invalid
URL values.

### DIFF
--- a/internal/cmd/root/products/konnect/loginKonnect.go
+++ b/internal/cmd/root/products/konnect/loginKonnect.go
@@ -9,6 +9,7 @@ import (
 	"github.com/kong/kongctl/internal/cmd"
 	"github.com/kong/kongctl/internal/cmd/root/products/konnect/common"
 	"github.com/kong/kongctl/internal/cmd/root/verbs"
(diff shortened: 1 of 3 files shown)


rspurgeon:
Thank you @Prashansa-K this looks good as I manually test.  I'm wondering if we could improve this case:
```
└─❯❯ ./kongctl login --base-auth-url https://konghq.com
Error: invalid character '<' looking for beginning of value
```
or
```
└─❯❯ ./kongctl login --base-auth-url https://get.konghq.com
Error: invalid character '<' looking for beginning of value
```

Prashansa-K:
@rspurgeon Added the fix for the error you found as well as a unit test case.

rspurgeon (approved):

Issue #731, opened 2026-04-02
Feat: Implement parallel planning and execution operations


rspurgeon:
The current planner and executor applies every change sequentially regardless of whether operations depend on each other. Creating 20 unrelated APIs takes 20 serial round trips.

`deck` and `terraform` both support a `--parallelism N` flag that runs independent operations concurrently.

* The execution order already encodes which operations must be sequenced. This makes adding parallelism for independent operations well-defined: anything not connected in the dependency graph can safely run concurrently.  
* Both unit and integration tests run with `go test -race` in CI, so any data races introduced by concurrent execution will be caught automatically.

harshadixit12:
## Additional Context from Preliminary Research

### 1. Terminology and Configuration

-   Use the term **concurrency** (not parallelism).

-   Expose a user-facing flag:

    ``` bash
    --max-concurrency <N>
    ```

-   This defines the **maximum number of operations that can execute
    concurrently**.

------------------------------------------------------------------------

### 2. Problem: Plan Is Not Concurrency-Safe

The current `plan` output cannot be safely executed concurrently.

#### Root Causes

Dependencies are **incomplete, implicit, and fragmented** across
multiple fields:

-   `execution_order`
    -   Implies sequencing but does **not guarantee independence**
-   `!ref` references
    -   Create implicit dependencies across steps
-   `depends_on`
    -   Explicit but not exhaustive
-   Parent-child relationships
-   `references` field

#### Resulting Issues

-   No **single source of truth** for dependencies
-   Hidden or indirect dependencies
-   Executor cannot reliably determine safe parallel execution
-   Risk of:
    -   race conditions
    -   invalid resource state
    -   non-deterministic execution

------------------------------------------------------------------------

### 3. Goal

Enable safe concurrent execution by ensuring:

-   **All dependencies are explicit and complete**
-   The plan can be used to derive **independent execution units**
-   Execution remains **deterministic regardless of concurrency level**

------------------------------------------------------------------------

### 4. Potential Approaches

#### Option A: Make Plan Concurrency-Aware (Preferred)

Enhance the planner to:

-   Construct a **fully resolved dependency graph**
-   Consolidate all dependency sources into a **single canonical
    representation**
-   Explicitly define **parallelizable groups of changes**

Example:

``` yaml
execution_groups:
  - [A]
  - [B, C]   # safe to execute concurrently
  - [D]
```

Executor responsibilities:

-   Respect ordering between groups
[...]

PR #967, opened 2026-05-03
feat: Add concurrent execution to declarative engine

harshadixit12:
## Summary
In this PR, we are enabling parallel execution of plan.
We are introducing a new flag `--max-concurrency` for `apply, sync, delete` commands, which defaults to 5 (based on konnect rate limits over a minute, and 200ms response time).

Implements https://github.com/Kong/kongctl/issues/731, and by improving on the plan in https://github.com/Kong/kongctl/issues/731#issuecomment-4380179046

The concurrent execution is achieved by:
1. Planner calculating execution groups, members of which are independent
2. Executor spinning up goroutines, and limiting their number using [errgroup.SetLimit()](https://pkg.go.dev/golang.org/x/sync/errgroup#Group.SetLimit)

The design can be read about here https://github.com/Kong/kongctl/pull/967/changes#diff-abc1bc172787ebd27c50dcbef28b4f4be77b407be2c97f3bc91de9da151ca251

## Benchmarks
This implementation: https://github.com/Kong/kongctl/actions/runs/25629149953

Screenshot 2026-05-10 at 6 38 53 PM

Existing implementation on main: https://github.com/Kong/kongctl/actions/runs/25629344884
Screenshot 2026-05-10 at 6 39 00 PM


internal/cmd/root/products/konnect/declarative/declarative.go
```diff
@@ -92,6 +92,12 @@ Defaults to each -f source root (file: its parent dir, dir: the directory itself
 - Config path: [ %s ]`, baseDirConfigPath))
 }
 
+func addParallelismFlag(cmd *cobra.Command) {
+	cmd.Flags().Int("parallelism", executor.DefaultParallelism,
+		fmt.Sprintf(`Maximum number of concurrent API operations during execution (min %d, max %d).`,
```
rspurgeon: @harshadixit12 I know this is WIP but I took a glance at it. One small thing is wording around flags. I think if a flag represents the maximum amount of something it should have that in the name. I propose:
```
--max-concurrency
```
where concurrency feels like the more precise term for I/O-bound API operations

internal/declarative/executor/executor.go
```diff
@@ -103,13 +107,26 @@ type Executor struct {
 	planBaseDir    string
 }
 
+// DefaultParallelism is the default number of concurrent executor operations.
+const DefaultParallelism = 10
+
+// MaxParallelism is the maximum allowed concurrent operations.
```
rspurgeon: As mentioned above and the comment suggests, I think `concurrent` is the better term for us to use broadly to describe this feature. My 2 cents

harshadixit12: Makes sense 👍

rspurgeon (changes requested):
Superseded by the follow-up inline review comments on the relevant diff lines. Keeping the requested-changes state, but the actionable findings are now attached to code references.

internal/declarative/executor/executor.go
```diff
@@ -108,6 +113,16 @@ type Executor struct {
 	planBaseDir        string
 }
 
+// DefaultMaxConcurrency is the default --max-concurrency value (sequential baseline).
+const DefaultMaxConcurrency = 10
```
rspurgeon: This default changes existing `apply`/`sync`/`delete` behavior from sequential to concurrent because the Cobra flag also uses `executor.DefaultMaxConcurrency`. The design in #731 explicitly calls for default concurrency of `1` to preserve behavior and avoid accidental API pressure; it also suggested a much smaller cap. Please make the default sequential and validate/cap user input explicitly.

internal/declarative/executor/executor.go
```diff
@@ -364,52 +386,365 @@ func NewWithOptions(client *state.Client, reporter ProgressReporter, dryRun bool
...
-			if plan.Changes[j].ID == changeID {
-				change = &plan.Changes[j]
-				break
+	// Choose execution strategy.
+	// If the plan contains execution_groups (i.e. it was produced by a concurrency-aware
+	// planner), use group-based concurrent execution.
+	// Otherwise fall back to the flat ExecutionOrder path (legacy plans).
+	if len(plan.ExecutionGroups) > 0 {
+		e.executeGroupsConcurrent(ctx, plan, result)
+	} else if e.concurrency > 1 {
```
rspurgeon: This fallback reintroduces the executor-derived dependency strategy that #731 rejected. Existing plans without `execution_groups` should execute sequentially via `execution_order`; otherwise an older plan file can run concurrently based on incomplete executor inference, especially with the current default above 1. Please only use concurrent execution when the planner authored `execution_groups`.

internal/declarative/planner/dependencies.go
```diff
@@ -100,6 +100,129 @@ func (d *DependencyResolver) ResolveDependencies(changes []PlannedChange) ([]str
...
+			}
+			allDepsPerNode[id][dep] = true
+			graph[dep] = append(graph[dep], id)
+			inDegree[id]++
+		}
+
+		for _, dep := range change.DependsOn {
+			addEdge(dep)
+		}
+		for _, dep := range d.findImplicitDependencies(change, changes) {
```
rspurgeon, line 157: This makes `execution_groups` depend on `findImplicitDependencies`, but that helper currently only treats `ID == "[unknown]"` as unresolved. Several planners use empty IDs for runtime-resolved refs, and some update paths add `__REF__` placeholders without setting `ID`, so the canonical dependency closure can still miss edges. Please use one unresolved-reference rule (`""` and `"[unknown]"`) and/or treat placeholders as dependencies regardless of `ID`, with planner tests for both forms.

internal/declarative/executor/executor.go
```diff
@@ -560,6 +900,7 @@ func (e *Executor) executeChange(ctx context.Context, result *ExecutionResult, c
 	}
```
rspurgeon, line 787: This lock protects result aggregation, but the block below still mutates `plan.Changes` to propagate created IDs. In the concurrent group path, workers hold pointers into `plan.Changes`, so any missed edge or future resource type can become a data race. #731 called for workers to operate on their own change copy plus a thread-safe runtime resolution context; runtime ID propagation should move out of the plan object.

internal/declarative/executor/executor.go
```diff
@@ -31,6 +33,9 @@ type Executor struct {
 	// Unified state cache
```
rspurgeon, line 35: Adding a mutex here only partially protects executor state. `stateCache` is still an unprotected set of maps, and concurrent portal/API child operations can populate or read those maps at the same time. The #731 design review called out the state cache as shared runtime state that needs protection; please guard it or make the cache concurrency-safe and add race tests for concurrent child resolution.

internal/declarative/executor/executor_test.go
```diff
@@ -455,3 +455,312 @@ func TestExecutor_ContinuesOnError(t *testing.T) {
 	// Verify all changes were attempted
 	assert.Len(t, reporter.CompleteChangeCalls, 3)
 }
+
+// // --- effectiveDependsOn tests ---
```
rspurgeon: These are the kinds of tests this feature needs, but they are commented out. As a result `make test -race` passes without exercising dependency-preserving concurrency, bounded execution, blocked dependents, CLI validation, or shared-state race safety. Please turn the relevant cases into active tests and add the planner/command coverage from the #731 test plan.

internal/cmd/root/products/konnect/declarative/declarative.go
```diff
@@ -2271,6 +2282,12 @@ func runSync(command *cobra.Command, args []string) error {
 	return nil
 }
 
+// maxConcurrencyFromCmd reads the --max-concurrency flag; returns 0 (use executor default) if not set.
+func maxConcurrencyFromCmd(cmd *cobra.Command) int {
+	v, _ := cmd.Flags().GetInt("max-concurrency")
```
rspurgeon: This reads the flag but never validates it at the command boundary. `--max-concurrency=0`, negative values, and values above the advertised max are silently mapped or clamped by `NewWithOptions`. Since the help text advertises min/max, the UX should fail fast with a clear error and command tests.

planning/declarative-parallel-execution-design.md
```diff
@@ -0,0 +1,222 @@
+# Declarative Parallel Execution Design
```
harshadixit12, line 1: This should be moved to /planning, will do.

internal/declarative/executor/executor.go
harshadixit12: Locks have been added for stateCache.

internal/declarative/executor/executor.go
harshadixit12: This intentionally not implemented - every `change` is executed in isolation, so there is no concurrent access. @rspurgeon please let me know if we want additional guardrails.

harshadixit12:
Another note to myself - add some benchmarks.

test/e2e/scenarios/control-plane/sync-groups/scenario.yaml
```diff
@@ -90,6 +90,7 @@ steps:
           - -f
           - "{{ .workdir }}/control-plane.yaml"
           - --auto-approve
+          - --max-concurrency=1
```
harshadixit12, line 93: This flag has been added to deletion of control planes because of a `conflict` returned by konnect - https://kongstrong.slack.com/archives/CQK8J4VN3/p1777983015046369

Without the flag, the deletion works on retry, and also deleting all control planes would be a rare event outside out e2e tests.

harshadixit12:
Another note: we already allow passing `decK` flags - `--parallelism` can be passed. https://developer.konghq.com/deck/gateway/apply/

internal/declarative/executor/executor.go
```diff
@@ -118,6 +124,18 @@ type Executor struct {
 	planBaseDir        string
 }
 
+// DefaultMaxConcurrency is the default --max-concurrency value.
```
harshadixit12, line 127: cc: @rspurgeon for visibility - I'm open to increasing the default.
The plan by codex had suggested capping the max concurrency at 20 - I think it's too conservative, I've picked 200 - open to change here as well.

rspurgeon: You can leave the max at 200 if you think that's reasonable. I don't think 200 would result in any beneficial performance improvement for users but as a configuration max that seems fine.

rspurgeon (approved):

PR #1000, opened 2026-05-07
Simplify: references to avoid hardcoding and confusion

harshadixit12:
Rather than hardcoding ID as "" and "[unknown]", refactoring to use constants, and functions for initialising the types. This will make the code more readable.


rspurgeon:
@harshadixit12 please close or revive this PR when you are able.

Issue #733, opened 2026-04-03
Feat: add privacy-first opt-in CLI telemetry


rspurgeon:
### Work In Progress!!

Add a privacy-first, in-process, opt-in telemetry capability to `kongctl` so
the team can understand real-world CLI usage at GA without collecting secrets,
resource identifiers, or other user-sensitive data.

The goal is not "more data". The goal is better product decisions with the
smallest, safest event set that still answers useful questions such as:

- which major command areas are actually used
- which workflows fail most often
- which versions/platforms are active in the field
- whether new features are being adopted after release

## Why This Matters

Without usage telemetry, feature and UX decisions are based on anecdote,
support noise, and internal assumptions. That is especially risky over time as
the team needs evidence for:

- which command groups deserve continued investment
- where command failures cluster
- which platforms and versions matter in practice
- whether declarative workflows are becoming the dominant usage pattern

## Product Principles

For a developer CLI, telemetry is only likely to be accepted if it is visibly
respectful. The issue should treat the following as non-negotiable:

- explicit opt-in, not opt-out
- no prompt in CI or other non-interactive contexts
- no credentials, tokens, org IDs, resource names, file contents, request
  bodies, or free-form command arguments
- local inspection and clear documentation of exactly what is sent
- best-effort delivery only: telemetry must never block command execution
- easy disable, reset, and status inspection

## Proposed UX

Successful developer CLIs usually make telemetry legible and reversible rather
than burying it in docs. A practical UX for `kongctl` would be:

1. On first interactive `kongctl login`, show a concise one-time prompt:
   - telemetry is off by default until the user chooses
   - explain exactly what is collected and what is never collected
   - link to docs
2. In CI / non-interactive sessions:
   - do not prompt on any command except `kongctl login` which should not be used in CI
   - remain disabled unless explicitly enabled by config or env
3. Provide explicit config and env overrides:
   - config path: `telemetry.enabled`
   - env override: `KONGCTL_TELEMETRY_ENABLED=true|false`
   - honor `DO_NOT_TRACK=1` as a hard disable

That pattern is generally better received than silently sending events, and it
avoids surprising enterprise users running scripted automation.

## What To Collect

Start with a minimal schema. Good initial fields are:

- anonymized install identifier generated locally after opt-in
- `kongctl` version
- OS and architecture
- command path, for example `apply`, `sync`, `login`, `get apis`
- high-level execution area, for example `declarative`, `konnect-imperative`,
  `auth`, `config`
[...]

PR #1022, opened 2026-05-11
Feat: add initial usage data collection and opt-out capabilities

Prashansa-K:
## Summary

Adds MVP command-level telemetry for `kongctl` so Kong can gauge CLI usage by
version, platform, and command area while keeping the event shape intentionally
small.

Telemetry is enabled by default and emits one best-effort event after a command
runs. Events are sent through the UDP sink when telemetry is enabled and debug
mode is disabled. When `telemetry.debug=true`, events are written only to the
local JSONL debug file at `$XDG_CONFIG_HOME/kongctl/logs/telemetry.log` (or
`$HOME/.config/kongctl/logs/telemetry.log`) and are not sent upstream.

`kongctl login` now uses the interactive login flow to disclose telemetry and
ask: `Allow kongctl to collect usage data on this device? [Y/n]:`. The answer
is saved to `$XDG_CONFIG_HOME/kongctl/.telemetry-enabled` (or
`$HOME/.config/kongctl/.telemetry-enabled`) and applies across local profiles
without writing to `config.yaml`.

## Event Shape

The current event contains only:

- schema version
- event timestamp
- `kongctl` version
- operating system and architecture
- command path, such as `get apis` or `apply`

The event does not include command arguments, flag values, auth tokens, request
or response bodies, resource names, resource IDs, configuration file contents,
file paths, usernames, email addresses, hostnames, or IP addresses.

## Opt Out

Telemetry can be disabled with any of these mechanisms:

- `kongctl --no-telemetry <command>` for one command invocation
- `KONGCTL_NO_TELEMETRY=true kongctl <command>` for the current process
- `DO_NOT_TRACK=1 kongctl <command>` using the cross-vendor convention
- `$XDG_CONFIG_HOME/kongctl/.telemetry-enabled` containing `false`
- profile config:

```yaml
default:
  telemetry:
    enabled: false
```

Precedence is: hard opt-outs (`--no-telemetry`, `DO_NOT_TRACK=1`, and
`KONGCTL_NO_TELEMETRY=true`), then the local `.telemetry-enabled` preference
file, then profile config, then the default enabled behavior. The local
preference file overrides profile config; an invalid preference file disables
telemetry conservatively and logs a warning.

Direct CI workflows that run `kongctl` set `DO_NOT_TRACK=1` so test invocations
do not emit telemetry.

## Validation

- `go test -race -count=1 ./internal/telemetry`
- `go test -race -count=1 ./internal/cmd/root/products/konnect`
- `go test -race -count=1 . ./internal/... ./pkg/... ./skills ./test/cmd ./test/config`
- `make build`
- `make lint`
- `git diff --check`

E2E tests were not run.


internal/telemetry/event.go
```diff
@@ -0,0 +1,24 @@
...
+
+	Version string `json:"version"`
+	OS      string `json:"os"`
+	Arch    string `json:"arch"`
+
+	CommandPath string   `json:"command_path"`
+	FlagsSet    []string `json:"flags_set,omitempty"`
+	ExecArea    string   `json:"exec_area"`
+
+	Outcome    string `json:"outcome,omitempty"`
```
Prashansa-K: I have kept the fields here with omitempty. I can comment them out too if necessary. At the moment, we are not capturing this information.

internal/telemetry/event.go
```diff
@@ -0,0 +1,24 @@
...
+
+import "time"
+
+// SchemaVersion identifies the shape of Event.
+const SchemaVersion = 1
+
+// Event is the single telemetry record kongctl emits per command execution.
+type Event struct {
+	SchemaVersion int       `json:"schema_version"`
+	Timestamp     time.Time `json:"timestamp"`
```
Prashansa-K, line 11: @rspurgeon Let me know if you think we would require a timestamp here or should we just rely on the ingestion timestamp, if a provider has that.

internal/telemetry/event.go
```diff
@@ -0,0 +1,24 @@
...
+
+	Version string `json:"version"`
+	OS      string `json:"os"`
+	Arch    string `json:"arch"`
+
+	CommandPath string   `json:"command_path"`
+	FlagsSet    []string `json:"flags_set,omitempty"`
```
Prashansa-K: Need your inputs here as well for flags - do we keep them or not? I have just used a allow-list for now so that the field doesn't blow up. I can switch to a block list as well.

internal/telemetry/outcome.go
```diff
@@ -0,0 +1,80 @@
+package telemetry
```
Prashansa-K: Keeping this here so that it can be hooked later on.

internal/telemetry/sink.go
```diff
@@ -0,0 +1,106 @@
...
+// The file is opened lazily on first Emit so a disabled sink imposes no IO.
+func NewFileSink(path string) Sink {
+	return &fileSink{path: path}
+}
+
+func (s *fileSink) Emit(_ context.Context, e Event) error {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+
+	f, err := os.OpenFile(s.path, os.O_APPEND|os.O_CREATE|os.O_WRONLY, 0o600)
```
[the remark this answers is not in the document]
Prashansa-K: @AikidoSec ignore: This feature is gated behind a config variable. The file is opened with os.O_WRONLY with restrictive permissions. The filename is a hardcoded constant ("telemetry.log"), and [cfg.GetPath()]returns the kongctl config file location — not user-supplied input.

internal/telemetry/event.go
rspurgeon: Let's go ahead and capture the client timestamp

internal/telemetry/event.go
rspurgeon: IMO for version 1 of this we keep flags off

internal/cmd/root/root.go
```diff
@@ -147,6 +151,18 @@ func newRootCmd() *cobra.Command {
 			ctx = context.WithValue(ctx, build.InfoKey, buildInfo)
 			ctx = context.WithValue(ctx, log.LoggerKey, logger)
 			ctx = theme.ContextWithPalette(ctx, theme.Current())
+
+			if telemetryRecorder == nil {
```
rspurgeon, line 212: This stores a single-use recorder in a package global, but `Execute` later closes it and never clears it. If `root.Execute` runs more than once in the same process with telemetry enabled, the second run reuses a recorder whose `events` channel was closed by the first run, so `Finalize` can panic with `send on closed channel`. The CLI entrypoint currently calls `Execute` once, but tests, embedding, or future command harnesses can reasonably exercise this path. Consider making the recorder per-`Execute`, or resetting `telemetryRecorder` after `Close`.

internal/telemetry/area.go
```diff
@@ -0,0 +1,32 @@
...
+)
+
+// AreaFor classifies a fully-qualified cobra command path
+// (e.g. "kongctl get orgs") into a high-level execution area.
+func AreaFor(commandPath string) string {
+	fields := strings.Fields(commandPath)
+	if len(fields) < 2 {
+		return AreaOther
+	}
+	switch fields[1] {
```
rspurgeon: This hard-coded verb switch misses several real root verbs, so their telemetry will be classified as `other`: `api`, `listen`, `lint`, `install`, extension verbs, `explain`, and `scaffold`. At least `api` and `listen` seem like they should not land in `other`, so this will skew telemetry once people start analyzing execution areas. It would be safer to centralize this around the existing verb constants or an explicit verb-to-area map that includes every root verb intentionally.

internal/telemetry/event.go
rspurgeon: These fields are part of schema version 1, but `Finalize` currently does not emit `Outcome`, `DurationMs`, or `Cancelled` values. That creates a stable schema contract before the data is ready, and consumers may interpret missing fields as unavailable rather than deliberately deferred. Consider either wiring these fields now or leaving them out of v1 until outcome reporting is actually enabled.

internal/telemetry/telemetry.go
```diff
@@ -0,0 +1,268 @@
...
+	}
+	return nil
+}
+
+func (r *Recorder) dispatch() {
+	defer close(r.done)
+	// The dispatcher outlives any request-scoped context: it must be able to
+	// drain in-flight events even when the original command context has been
+	// cancelled. Using a fresh background context here is intentional.
+	ctx := context.Background()
```
rspurgeon: Current sinks are local/no-op, but this dispatcher context is unbounded. When the TODO HTTP/network sink lands, the sink itself will need to enforce a short request timeout; otherwise `Close` returns after `flushTimeout`, but the dispatcher goroutine can remain blocked until process exit. It would be worth documenting this requirement on the `Sink` contract or passing a bounded context into sink operations from the dispatcher.

rspurgeon (commented):
I reviewed the latest pushed version of this telemetry PR. The earlier file-close/docstring/test coverage comments look addressed; I left the remaining actionable feedback inline.

internal/telemetry/area.go
Prashansa-K: Got it.

Would `api` be considered imperative? 
I can add listen under `AreaAuditLogs` maybe.

rspurgeon: Yes, despite not being a verb `api` is used to invoke single api requests

internal/cmd/root/root.go
Prashansa-K: Addressed this.

internal/telemetry/udp_sink.go
```diff
@@ -0,0 +1,174 @@
...
+	// receiver; `"` and \n would break the quoted form itself.
+	const quoteTriggers = " ;\"\n"
+
+	// Empty values are quoted so the receiver sees `key=""` instead of a
+	// bare `key=`, which key=value parsers can read ambiguously (either as
+	// an empty value or as a value that runs into the next pair).
+	if v == "" {
+		return true
+	}
+	return strings.ContainsAny(v, quoteTriggers)
```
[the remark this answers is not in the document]
Prashansa-K: Addressed.

Prashansa-K:
@rspurgeon I have added a bunch of changes after your last review. 
- I have addressed the comments you gave + the copilot comments. 
- Added a udp sink so that we can pipe the analytics to splunk.
- I also added an env variable ~KONGCTL_TELEMETRY_ENABLED~ KONGCTL_NO_TELEMETRY which users can choose to opt out of telemetry completely, independent of profiles. This would work for CI environments as well.
- Also, added DO_NOT_TRACK support as per conventions. If set to 1, telemetry won't be sent. I have added this to our CI workflows.
- Enabled the telemetry by default.

.github/workflows/agentics-maintenance.yml
```diff
@@ -12,7 +12,7 @@
 # \  /\  / (_) | | | | ( | | | | (_) \ V  V /\__ \
```
Prashansa-K, line 12: Got this from `gh aw compile` command.

internal/telemetry/telemetry.go
```diff
@@ -0,0 +1,485 @@
...
+	if err != nil {
+		return err
+	}
+	if err := os.MkdirAll(filepath.Dir(path), 0o755); err != nil {
+		return err
+	}
+	return os.WriteFile(path, []byte(strconv.FormatBool(enabled)+"\n"), 0o600)
+}
+
+type isSetConfig interface {
```
rspurgeon: **High:** `IsSet` was added to the concrete `ProfiledConfig` only (not to the `config.Hook` interface), so this file has to define a private `isSetConfig` interface and do a runtime type assertion (`cfg.(isSetConfig)`). Two problems:

1. Any future `Hook` implementation that omits `IsSet` silently loses the conflict-detection warning — no compile error, no test failure, just quietly degraded behavior. The current test fakes happen to implement it, so the production path is exercised; that's lucky, not enforced.
2. Naming collides with viper's own `IsSet`, which has very different semantics (env + flag + override + config). `ProfiledConfig.IsSet` is file-only via `InConfig`. Future readers will misread it.

Suggestion: lift `IsSet` onto `config.Hook` (with a clearly documented "config-file-only" contract) or move the conflict-logging concern out of the telemetry package entirely.

internal/cmd/root/products/konnect/loginKonnect.go
```diff
@@ -73,7 +84,130 @@ func displayUserInstructions(resp auth.DeviceCodeResponse) {
...
+	interrupt <-chan os.Signal,
+) (bool, bool, error) {
+	if ctx == nil {
+		ctx = context.Background()
+	}
+	reader := bufio.NewReader(in)
+	for attempt := range 2 {
+		lineCh := make(chan string, 1)
+		errCh := make(chan error, 1)
+		go func() {
```
rspurgeon, line 171: **High:** this goroutine calls `reader.ReadString('\n')`, which is uncancellable. When the parent select hits `<-ctx.Done()` or `<-interrupt` (lines 179, 181), we return — but the goroutine is still parked on `read(2)` against stdin. There's no way to interrupt a blocking stdin read in Go.

In the `login` flow this is probably benign because the process exits shortly after, but:
- It outlives the `defer signal.Stop(sigCh)` in `promptTelemetryPreference`.
- It will trip `go test -race`/goroutine-leak detectors if anyone wires those up later.
- If `handleTelemetryPreference` is ever reused outside this short-lived `login` flow, the leak becomes real.

At minimum: add a comment acknowledging it. Better: convert stdin to non-blocking, or accept fire-and-forget semantics explicitly.

internal/cmd/root/root.go
```diff
@@ -652,6 +684,11 @@ func Execute(ctx context.Context, s *iostreams.IOStreams, bi *build.Info) {
 	if err == nil {
 		executedCmd, err = rootCmd.ExecuteContextC(ctx)
 	}
+	if telemetryRecorder != nil {
```
rspurgeon: **High:** `telemetryRecorder` is a package-level global (declared at root.go:101) and is cleaned up here only on the happy path. A panic anywhere between `NewRecorder` in the PersistentPreRunE (root.go:213) and reaching this block will leak the dispatcher goroutine and any open UDP/file resources.

The rest of this file uses globals for `logger`, `logFile`, etc., so this matches local style — but for the telemetry recorder specifically, please consider:
- A `defer rec.Close(ctx)` registered immediately after construction in the PersistentPreRunE, or
- Wrapping Execute's body in a `defer func() { if telemetryRecorder != nil { ... } }()` so the cleanup runs even on panic.

Also: this makes `Execute` non-reentrant, which matters if anything ever wants to call it from tests.

internal/telemetry/recorder_test.go
```diff
@@ -0,0 +1,555 @@
...
+
+func TestContextWithRecorder_RoundTrip(t *testing.T) {
+	rec := &Recorder{}
+	ctx := ContextWithRecorder(t.Context(), rec)
+	if got := FromContext(ctx); got != rec {
+		t.Errorf("FromContext after ContextWithRecorder = %v, want %v", got, rec)
+	}
+}
+
+// We can't use these tests in CI as we use DO_NOT_TRACK=1 to disable telemetry in CI,
```
rspurgeon: **Medium:** the comment block says these tests can't run in CI because of `DO_NOT_TRACK=1`, but the active tests just below (e.g. `TestNewRecorder_PreferenceFileOverridesConfig` at line 195, `TestNewRecorder_InvalidPreferenceFileDisables` at line 234) handle exactly that with `t.Setenv(EnvDoNotTrack, "0")` and `t.Setenv(EnvNoTelemetry, "false")`. The same fix applies here.

This is the kind of "half-finished implementation" `AGENTS.md` calls out — either fix the three tests and uncomment them, or delete them. As-is, ~30 lines of comment-only code will rot.

internal/telemetry/event.go
```diff
@@ -0,0 +1,23 @@
...
+	SchemaVersion int       `json:"schema_version"`
+	Timestamp     time.Time `json:"timestamp"`
+
+	Version string `json:"version"`
+	OS      string `json:"os"`
+	Arch    string `json:"arch"`
+
+	CommandPath string `json:"command_path"`
+
+	// TODO: Uncomment when we start reporting outcomes.
```
rspurgeon: **Medium:** these commented-out fields (along with the `Categorize`/`Outcome` machinery in `outcome.go` and the commented blocks in `telemetry.go:386-392` and `telemetry.go:471-474`) are dead-on-arrival until someone uncomments them. That's ~150 lines (including the 70-line `outcome_test.go`) of code that exists only to support a path that's currently switched off.

This cuts against the repo guideline "Don't add abstractions beyond what the task requires." Two cleaner options:

1. Wire outcomes in now — the assignment is one line, `Categorize` is already tested, this finishes the schema.
2. Delete `outcome.go`, `outcome_test.go`, and the commented blocks until you're ready. Bring them back in a follow-up PR.

The current state is the worst of both worlds: code that has to be kept compiling and reviewed but isn't actually doing anything.

internal/telemetry/telemetry.go
```diff
@@ -0,0 +1,485 @@
...
+type Recorder struct {
+	enabled bool
+	logger  *slog.Logger
+
+	cfg config.Hook
+
+	staticEvent Event // pre-populated Version/OS/Arch
+
+	mu        sync.Mutex
+	startedAt time.Time
```
rspurgeon: **Medium:** `startedAt` is written by `Begin` but never read — the duration calculation is in the commented block in `Finalize` (lines 386-392). Either uncomment that block (preferred — Begin/end timing is cheap and useful) or drop `Begin` and `startedAt` until the duration is actually emitted. Same complaint as the commented `Event` fields.

internal/cmd/root/products/konnect/loginKonnect.go
```diff
@@ -73,7 +84,130 @@ func displayUserInstructions(resp auth.DeviceCodeResponse) {
...
+Collected:
+  - kongctl version
+  - operating system and architecture
+  - command path, such as "login" or "get apis"
+
+Not collected:
+  - command arguments or flag values
+  - resource names or IDs
+  - auth tokens, request bodies, or response bodies
+  - config file contents, file paths, hostnames, usernames, or email addresses
```
rspurgeon, line 134: **Medium:** this disclosure's "Not collected" list ends with `config file contents, file paths, hostnames, usernames, or email addresses`. The README's equivalent section (README.md:140-145) splits those into two lines and orders the bullets differently. A user who reads both will reasonably ask which one is authoritative.

Recommendation: define the disclosure copy in one place (a single Go constant or a `//go:embed`'d markdown snippet) and reference it from both the CLI and the README, or at minimum keep the strings byte-identical via review discipline.

internal/cmd/root/products/konnect/loginKonnect.go
```diff
@@ -73,7 +84,130 @@ func displayUserInstructions(resp auth.DeviceCodeResponse) {
...
+	streams *iostreams.IOStreams,
+	cfg config.Hook,
+	rec *telemetry.Recorder,
+) error {
+	if streams == nil || cfg == nil || rec == nil || !rec.Enabled() || telemetry.PreferenceFileExists(cfg) {
+		return nil
+	}
+
+	if !loginInputIsTerminal(streams.In) {
+		writeTelemetryDisclosure(streams.Out, false)
```
rspurgeon: **Medium:** in the non-interactive case the full disclosure is written to `streams.Out` (stdout). For scripted `kongctl login` usage that captures stdout, every run will inject the disclosure into the captured stream — and because no preference file ever gets written in this branch, it never stops. Conventional fix is to send notices to `streams.ErrOut` for non-interactive callers.

(Today `login` is interactive-only, so the impact is limited, but still worth changing now.)

internal/cmd/root/products/konnect/loginKonnect.go
```diff
@@ -73,7 +84,130 @@ func displayUserInstructions(resp auth.DeviceCodeResponse) {
...
+
+		var line string
+		select {
+		case <-ctx.Done():
+			return false, false, ctx.Err()
+		case <-interrupt:
+			fmt.Fprintln(out)
+			return false, false, context.Canceled
+		case err := <-errCh:
+			_ = err
```
rspurgeon: **Nit:** `case err := <-errCh: _ = err` can be `case <-errCh:` — there's no need to bind. Linter likely flags the unused name on its own.

internal/telemetry/telemetry.go
```diff
@@ -0,0 +1,485 @@
...
+// }
+
+func loggerOrDiscard(l *slog.Logger) *slog.Logger {
+	if l != nil {
+		return l
+	}
+	return slog.New(slog.NewTextHandler(discardWriter{}, nil))
+}
+
+type discardWriter struct{}
```
rspurgeon: **Nit:** `discardWriter` + its `Write` method can be replaced with `io.Discard`:

```go
return slog.New(slog.NewTextHandler(io.Discard, nil))
```

Then delete `type discardWriter struct{}` and the `Write` method below.

internal/telemetry/outcome.go
```diff
@@ -0,0 +1,80 @@
...
+	if errors.As(err, &urlErr) {
+		return OutcomeNetworkError
+	}
+
+	var netErr net.Error
+	if errors.As(err, &netErr) {
+		return OutcomeNetworkError
+	}
+
+	var execErr *cmd.ExecutionError
```
rspurgeon: **Nit:** this branch (`*cmd.ExecutionError` -> `OutcomeInternalError`) returns the same value as the fallthrough at line 79. The doc comment at line 38 lists them as separate bands. Either drop this branch entirely, or have it return a distinct outcome to match the doc. Currently it's a no-op.

internal/telemetry/udp_sink.go
```diff
@@ -0,0 +1,174 @@
...
+	"strconv"
+	"strings"
+	"sync"
+
+	"github.com/kong/kongctl/internal/meta"
+)
+
+const (
+	signalName   = meta.CLIName
+	reportsAddr  = "kong-hf.konghq.com:61829"
```
rspurgeon, line 22: **Nit:** `reportsAddr` is hardcoded with no env-var or config override. `telemetry.debug` is the only escape hatch and it only redirects to a file — there's no way to point production builds at a staging UDP receiver. Fine for v1, but worth a TODO so future plumbing is straightforward.

internal/telemetry/telemetry.go
```diff
@@ -0,0 +1,485 @@
...
+	return err == nil
+}
+
+func ReadPreference(cfg config.Hook, logger *slog.Logger) (bool, bool) {
+	path, err := PreferenceFilePath(cfg)
+	if err != nil {
+		return false, false
+	}
+	data, err := os.ReadFile(path)
+	if err != nil {
```
rspurgeon, line 135: **Low (privacy-adjacent):** if the preference file exists but is unreadable for any reason other than `IsNotExist` (transient FS error, permissions glitch after `chown`, etc.), we log a warning and `return false, false`, which makes `resolveEnabled` fall through to `cfg.GetBool(ConfigKeyEnabled)` — defaulting to enabled.

A user who deliberately opted out via the file could be silently re-enabled. Privacy-safe default would be to fail closed: treat any non-`IsNotExist` read error as "opted out." Same logic could apply to the invalid-content branch at line 142, which already returns `(false, true)` — good — so this is just about the unreadable case.

README.md
```diff
@@ -179,6 +232,7 @@ KONGCTL_DEFAULT_OUTPUT=yaml kongctl get apis
 To clear saved Konnect credentials for a profile, run `kongctl logout [--profile <name>]`. This removes the local device
 flow token file so that subsequent commands prompt you to authenticate again with `kongctl login`.
 
+
```
rspurgeon: **Nit:** extra blank line — there are now two blank lines between the `kongctl logout` paragraph and `### Authentication Options`. Likely accidental.

rspurgeon (commented):
Review of the telemetry MVP. Findings are posted as inline comments below, grouped by severity in this summary. Nothing here is blocking — flagging things for discussion.

**High**
1. `IsSet` is added to `ProfiledConfig` but not to the `Hook` interface, so telemetry does a private type assertion to reach it. Brittle; future `Hook` impls will silently lose the conflict warning.
2. Goroutine reading stdin in `readTelemetryPreferenceAnswer` leaks on `ctx.Done()` / Ctrl-C. Probably benign because the process exits soon after, but worth at least a comment.
3. `telemetryRecorder` lives at package scope in `root.go` and is cleaned up only on the happy path in `Execute`. A panic between `NewRecorder` and the cleanup block leaks the dispatcher goroutine.

**Medium**
4. Commented-out test functions in `recorder_test.go` should be either fixed (they just need `t.Setenv(EnvDoNotTrack, "0")` like the active tests) or deleted.
5. The outcome plumbing (`outcome.go`, `Categorize`, the commented `Event` fields, the commented `isCanceled` helper) is dead code until you flip the TODOs. Either wire it up now or remove until you're ready.
6. `Recorder.Begin` stores `startedAt`, but `Finalize` never reads it (the duration computation is in the same commented block as #5).
7. README and CLI disclosure differ on which fields are listed under "Not collected". They should match.
8. Non-interactive disclosure writes to stdout, polluting any captured output. Probably should be stderr.

**Low / nits**
9. Stray extra blank line in README before `### Authentication Options`.
10. `case err := <-errCh: _ = err` — just `case <-errCh:`.
11. Hand-rolled `discardWriter` where `io.Discard` would do.
12. `Categorize`'s `*cmd.ExecutionError` branch returns the same value as the fallthrough — either drop it or differentiate.
13. Hardcoded UDP `reportsAddr`; no override for staging. Worth a TODO.
14. Preference-file read errors (non-`IsNotExist`) silently fall through to config, which can re-enable telemetry. Privacy-safe default would be to fail closed.
15. (nit) `trimBinaryPrefix` correctly leaves `kongctl-extension` alone because of the trailing-space check — worth a one-line comment so the test's intent survives a future refactor.

What looks good: command_path filtering is verified by a dedicated test, bounded Close flush + ctx-aware dial gives a sound shutdown story, DO_NOT_TRACK strict `"1"` parsing matches spec, file-mode `0o600` is appropriate for both files.

internal/telemetry/outcome.go
rspurgeon: Removed in fd3cc0c7 to keep the MVP schema minimal. We can bring outcome categorization back in a follow-up when it is wired into emitted events.

.github/workflows/agentics-maintenance.yml
rspurgeon: The gh aw workflow churn was backed out in 91b8d424, so this thread is now outdated.

internal/telemetry/telemetry.go
rspurgeon: Addressed in fd3cc0c7: added config.Hook.InConfig with a config-file-only contract and removed the private runtime type assertion from telemetry.

internal/cmd/root/products/konnect/loginKonnect.go
rspurgeon: Addressed in fd3cc0c7: added an explicit comment documenting the fire-and-forget goroutine semantics for the uncancellable terminal line read.

internal/cmd/root/root.go
rspurgeon: Addressed in fd3cc0c7: Execute now funnels recorder cleanup through cleanupTelemetryRecorder, resets the global, and has a panic defer that closes telemetry resources before re-panicking.

internal/telemetry/recorder_test.go
rspurgeon: Addressed in fd3cc0c7: uncommented the recorder tests and made them CI-safe with explicit telemetry env overrides.

internal/telemetry/event.go
rspurgeon: Addressed in fd3cc0c7: removed the commented event fields, outcome.go, outcome_test.go, and the commented outcome blocks from telemetry.go.

internal/telemetry/telemetry.go
rspurgeon: Addressed in fd3cc0c7: removed Begin and startedAt while duration is not part of the emitted MVP event.

internal/cmd/root/products/konnect/loginKonnect.go
rspurgeon: Addressed in fd3cc0c7: the README Not collected list now matches the CLI disclosure wording and order.

internal/cmd/root/products/konnect/loginKonnect.go
rspurgeon: Addressed in fd3cc0c7: non-interactive disclosure now writes to ErrOut instead of stdout, with the test updated to verify stdout stays clean.

internal/cmd/root/products/konnect/loginKonnect.go
rspurgeon: Addressed in fd3cc0c7: changed the select arm to case <-errCh:.

internal/telemetry/telemetry.go
rspurgeon: Addressed in fd3cc0c7: replaced discardWriter with io.Discard.

internal/telemetry/outcome.go
rspurgeon: Addressed in fd3cc0c7 by deleting the dormant outcome machinery for the MVP.

internal/telemetry/udp_sink.go
rspurgeon: Addressed in fd3cc0c7: added a TODO beside reportsAddr for a future staging/config override.

internal/telemetry/telemetry.go
rspurgeon: Addressed in fd3cc0c7: non-IsNotExist preference-file read errors now fail closed, and there is a regression test for that path.

README.md
rspurgeon: Addressed in fd3cc0c7: removed the extra README blank line.

rspurgeon:
Also addressed the summary-only `trimBinaryPrefix` note in d94b825e by documenting why the prefix trim requires a trailing space, so extension binary names such as `kongctl-extension` remain untouched.

internal/telemetry/sink.go
```diff
@@ -0,0 +1,133 @@
...
+	s.mu.Lock()
+	defer s.mu.Unlock()
+
+	if dir := filepath.Dir(s.path); dir != "" && dir != "." {
+		if err := os.MkdirAll(dir, 0o755); err != nil {
+			return err
+		}
+	}
+
+	f, err := os.OpenFile(s.path, os.O_APPEND|os.O_CREATE|os.O_WRONLY, 0o600)
```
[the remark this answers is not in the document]
rspurgeon: @AikidoSec ignore:

What the code actually does:

  - The flagged line is os.OpenFile(s.path, os.O_APPEND|os.O_CREATE|os.O_WRONLY, 0o600) in internal/telemetry/sink.go.
  - That opens a telemetry debug log for append/write only, not read.
  - The filename is not user-supplied directly. It is built as:

  filepath.Join(filepath.Dir(cfg.GetPath()), "logs", "telemetry.log")

  - This path is only used when telemetry.debug=true; default is false.
  - Normal telemetry uses the UDP sink, not this file sink.

  Risk assessment:

  - Not a file inclusion/read vulnerability: there is no ReadFile or read of attacker-controlled file content at this sink.
  - Low local file-write concern only in debug mode: if a user chooses a custom config path, the debug telemetry log is written under that config file’s directory at logs/telemetry.log.
  - Possible symlink caveat: os.OpenFile follows symlinks. If an attacker can control the config directory and convince a more privileged user/process to run kongctl with
    telemetry.debug=true, they might redirect the append to another file via symlink. That is a broader local privilege/config-dir trust issue, not a remote/file-inclusion issue.

rspurgeon:
Closed #733

PR #1032, opened 2026-05-12
feat: telemetry info on login

Prashansa-K:
With this change, we will explicitly let the user know about the telemetry addition and
how to opt out if necessary.

Here's how it looks like:
Screenshot 2026-05-12 at 9 31 06 PM

Keeping this PR up if it is required in future.


rspurgeon:
Closing this in favor of adding directly to PR #1022. Additional changes there were significant enough to warrant a rework of this behavior.

Issue #732, opened 2026-04-02
Feat: add configurable retry and backoff for declarative Konnect operations


rspurgeon:
## Summary

Add a first-class retry/backoff capability for Konnect-backed CLI operations,
starting with the declarative runtime. Today, large `apply` / `sync` / `delete`
runs can fail hard on transient Konnect API errors or throttling even though
the underlying SDK already supports retry behavior that `kongctl` does not
currently enable.

## Problem

Current behavior has a few important gaps:

- Declarative Konnect calls use the shared SDK client, but the shared client
  does not enable SDK retry/backoff.
- The shared HTTP client has a 60 second timeout and transport tuning knobs,
  but those are not request retry controls.
- A merged API publication helper uses raw HTTP and bypasses generated SDK
  retry handling entirely.
- Recent retry/backoff work improved the e2e harness, not the production
  declarative CLI runtime.

This means transient failures such as `429`, `500`, `502`, `503`, `504`, and
selected connection-level I/O failures can still surface as user-visible hard
failures during declarative Konnect workflows.

## Desired Capability

`kongctl` should be able to automatically recover from transient Konnect API
failures with bounded retry/backoff behavior instead of requiring a manual
rerun for common throttling and temporary service issues.

At a minimum, the CLI should:

- retry `429`, `500`, `502`, `503`, and `504` for Konnect-backed operations
- honor `Retry-After` when Konnect provides it
- use exponential backoff with jitter
- optionally retry retryable connection errors for safe/idempotent requests
- preserve existing partial-success reporting once retries are exhausted
- emit trace/debug logs that make retry behavior visible to operators

## Proposed UX

Follow the existing `konnect.http-*` naming pattern and make the behavior
configurable both by flags and profile config.

Suggested settings:

- `--http-retry-max-attempts`
  Config path: `konnect.http-retry-max-attempts`
- `--http-retry-initial-interval`
  Config path: `konnect.http-retry-initial-interval`
- `--http-retry-max-interval`
  Config path: `konnect.http-retry-max-interval`
- `--http-retry-backoff-factor`
  Config path: `konnect.http-retry-backoff-factor`
- `--http-retry-jitter`
  Config path: `konnect.http-retry-jitter`
- `--http-retry-on-connection-errors`
  Config path: `konnect.http-retry-on-connection-errors`

Expected behavior:

- Conservative defaults should work out of the box for declarative Konnect
  commands without requiring users to tune retry behavior manually.
[...]

PR #1047, opened 2026-05-13
Feat: Add retry and backoff support for Konnect APIs in declarative configuration engine

harshadixit12:
Implements https://github.com/Kong/kongctl/issues/732

## Summary
Adds support for retries with backoff for **declarative engine** - `plan, diff, apply, sync, and delete` commands.
Since we want retry options to be configurable, we will use the request level overrides supported by speakeasy - https://www.speakeasy.com/docs/sdks/customize/runtime/retries#request-level-override
https://github.com/Kong/sdk-konnect-go#retries

The SDK supports retrying on some HTTP codes by default, and does not let us override them:
Supported HTTP codes: `429, 500, 502, 503, 504` ([source](https://github.com/Kong/sdk-konnect-go/blob/main/eventgatewaybackendclusters.go#L114))

SDK also implements jitter - https://github.com/Kong/sdk-konnect-go/blob/main/internal/utils/retries.go

## Default configuration
```
        http-retry-max-attempts: 5
        http-retry-initial-interval: 500
        http-retry-max-interval: 10000 #10s 
        http-retry-backoff-factor: 2.0
        http-retry-on-connection-errors: false
```

### Flags
```
--http-retry-max-attempts (int)
--http-retry-initial-interval (int - ms)
--http-retry-max-interval (int - ms)
--http-retry-backoff-factor (float)
--http-retry-on-connection-errors (bool)
```

### Using config file
```
myprofile:
  konnect:
        http-retry-max-attempts: 4
        http-retry-initial-interval: 500
        http-retry-max-interval: 120000 #2 mins
        http-retry-backoff-factor: 1.5
        http-retry-on-connection-errors: true
```

### To disable
```
--http-retry-max-attempts 1 //flag

// config-file
myprofile:
  konnect:
        http-retry-max-attempts: 1
```

## What is not covered
- Respecting `Retry-After` header - we have not declared this in our OAS from which the sdk is generated
- Additional custom logging for retries - we are using inbuild SDK feature, and have little control

### DIFF
--- a/internal/cmd/root/products/konnect/common/common.go
+++ b/internal/cmd/root/products/konnect/common/common.go
@@ -203,6 +203,9 @@ func ResolveRetryConfig(cfg config.Hook) (httpclient.RetryConfig, error) {
 	if err != nil {
 		return httpclient.RetryConfig{}, err
 	}
+	if factor == 0 {
+		factor = httpclient.DefaultRetryBackoffFactor
+	}
 	if factor <= 0 {
 		return httpclient.RetryConfig{}, fmt.Errorf("invalid %s value %f: must be >= 1",
 			HTTPRetryBackoffFactorConfigPath, factor)
--- a/internal/cmd/root/products/konnect/common/common.go
+++ b/internal/cmd/root/products/konnect/common/common.go
@@ -179,24 +179,42 @@ func ResolveRetryConfig(cfg config.Hook) (httpclient.RetryConfig, error) {
 	if err != nil {
 		return httpclient.RetryConfig{}, err
 	}
-	if initialIntervalMS < 0 {
-		return httpclient.RetryConfig{}, fmt.Errorf("invalid %s value %d: must be >= 0",
-			HTTPRetryInitialIntervalConfigPath, initialIntervalMS)
-	}
 	if initialIntervalMS == 0 {
 		initialIntervalMS = httpclient.DefaultRetryInitialIntervalMS
+	}
+	if initialIntervalMS < httpclient.MinRetryInitialIntervalMS {
+		return httpclient.RetryConfig{}, fmt.Errorf(
+			"invalid %s value %d: must be >= %d ms",
+			HTTPRetryInitialIntervalConfigPath, initialIntervalMS, httpclient.MinRetryInitialIntervalMS)
+	}
+	if initialIntervalMS > httpclient.MaxRetryInitialIntervalMS {
+		return httpclient.RetryConfig{}, fmt.Errorf(
+			"invalid %s value %d: must be <= %d ms",
+			HTTPRetryInitialIntervalConfigPath, initialIntervalMS, httpclient.MaxRetryInitialIntervalMS)
 	}
 
 	maxIntervalMS, err := resolveOptionalInt(cfg, HTTPRetryMaxIntervalConfigPath)
 	if err != nil {
 		return httpclient.RetryConfig{}, err
 	}
-	if maxIntervalMS < 0 {
-		return httpclient.RetryConfig{}, fmt.Errorf("invalid %s value %d: must be >= 0",
-			HTTPRetryMaxIntervalConfigPath, maxIntervalMS)
-	}
 	if maxIntervalMS == 0 {
 		maxIntervalMS = httpclient.DefaultRetryMaxIntervalMS
+	}
+	if maxIntervalMS < httpclient.MinRetryMaxIntervalMS {
+		return httpclient.RetryConfig{}, fmt.Errorf(
+			"invalid %s value %d: must be >= %d ms",
+			HTTPRetryMaxIntervalConfigPath, maxIntervalMS, httpclient.MinRetryMaxIntervalMS)
+	}
+	if maxIntervalMS > httpclient.MaxRetryMaxIntervalMS {
+		return httpclient.RetryConfig{}, fmt.Errorf(
+			"invalid %s value %d: must be <= %d ms",
+			HTTPRetryMaxIntervalConfigPath, maxIntervalMS, httpclient.MaxRetryMaxIntervalMS)
+	}
+	if initialIntervalMS > maxIntervalMS {
+		return httpclient.RetryConfig{}, fmt.Errorf(
+			"invalid configuration: %s (%d ms) must be <= %s (%d ms)",
+			HTTPRetryInitialIntervalConfigPath, initialIntervalMS,
+			HTTPRetryMaxIntervalConfigPath, maxIntervalMS)
 	}
 
 	factor, err := resolveOptionalFloat64(cfg, HTTPRetryBackoffFactorConfigPath)
--- a/internal/cmd/root/products/konnect/common/common.go
+++ b/internal/cmd/root/products/konnect/common/common.go
@@ -62,6 +62,12 @@ var (
 	RequestPageSizeConfigPath = "konnect." + RequestPageSizeFlagName
 
 	RegionConfigPath = "konnect." + RegionFlagName
+
+	HTTPRetryMaxAttemptsConfigPath        = "konnect." + cmdcommon.HTTPRetryMaxAttemptsConfigPath
+	HTTPRetryInitialIntervalConfigPath    = "konnect." + cmdcommon.HTTPRetryInitialIntervalConfigPath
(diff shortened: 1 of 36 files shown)


internal/konnect/httpclient/retry_config.go
```diff
@@ -0,0 +1,56 @@
...
+	// DefaultRetryBackoffFactor is the default exponential backoff multiplier.
+	DefaultRetryBackoffFactor = 2.0
+)
+
+// RetryConfig holds retry/backoff parameters resolved from flags/config.
+// All time intervals are stored as milliseconds. MaxAttempts is total
+// attempts including the first request.
+type RetryConfig struct {
+	Strategy              string
+	MaxAttempts           int
```
rspurgeon, line 45: This models the user-facing knob as a max attempt count, but the SDK retry config has no attempt cap. `ToSDKRetryConfig` converts this into `MaxElapsedTime`, so `--http-retry-max-attempts=5` can still produce more or fewer than 5 total requests depending on response timing, Retry-After, and backoff intervals. We should either expose an elapsed-time setting that matches the SDK semantics, or add our own attempt-limiting wrapper/test so the flag does what it says.

internal/cmd/root/products/konnect/common/common.go
```diff
@@ -143,6 +150,65 @@ func ResolveHTTPTransportOptions(cfg config.Hook) (httpclient.TransportOptions,
...
+func ResolveRetryConfig(cfg config.Hook) (httpclient.RetryConfig, error) {
+	maxAttempts, err := resolveOptionalInt(cfg, HTTPRetryMaxAttemptsConfigPath)
+	if err != nil {
+		return httpclient.RetryConfig{}, err
+	}
+	if maxAttempts < 0 {
+		return httpclient.RetryConfig{}, fmt.Errorf("invalid %s value %d: must be >= 0",
+			HTTPRetryMaxAttemptsConfigPath, maxAttempts)
+	}
+	if maxAttempts == 0 {
```
rspurgeon, line 169: Leaving retry config unset currently resolves to a backoff strategy and every declarative state call receives `WithRetries`. That means create/update/delete operations will retry 429/5xx by default. For mutating calls, an ambiguous 500/503 after the server has applied the write can duplicate creates or replay partial updates. Please make this behavior explicit and tested, or restrict default retries to read/idempotent operations and require opt-in for mutating calls.

internal/cmd/root/products/konnect/common/common.go
```diff
@@ -143,6 +150,65 @@ func ResolveHTTPTransportOptions(cfg config.Hook) (httpclient.TransportOptions,
...
+
+	maxIntervalMS, err := resolveOptionalInt(cfg, HTTPRetryMaxIntervalConfigPath)
+	if err != nil {
+		return httpclient.RetryConfig{}, err
+	}
+	if maxIntervalMS == 0 {
+		maxIntervalMS = httpclient.DefaultRetryMaxIntervalMS
+	}
+
+	factor, err := resolveOptionalFloat64(cfg, HTTPRetryBackoffFactorConfigPath)
```
rspurgeon, line 220: `strconv.ParseFloat` accepts values like `NaN` and `+Inf`, and the current `factor <= 0` check silently converts zero/negative values to the default. These should be rejected as invalid configuration instead of being accepted or normalized, otherwise we can pass non-finite values into the SDK backoff calculation.

internal/declarative/state/client.go
```diff
@@ -1877,7 +1889,7 @@ func (c *Client) CreateAPIPublication(
 		APIPublication: publication,
 	}
 
-	resp, err := c.apiPublicationAPI.PublishAPIToPortal(ctx, req)
+	resp, err := c.apiPublicationAPI.PublishAPIToPortal(ctx, req, c.retryOpts...)
```
rspurgeon, line 2044: This passes retry options to `PublishAPIToPortal`, but that helper has a merged-payload raw HTTP path that is selected for partial publication payloads and currently drops the options before calling `apiutil.Request*`. As a result, retry behavior for API publications depends on payload shape. Please either forward/apply the retry policy in that raw path or document/test that this path intentionally bypasses retry settings.

rspurgeon (commented):
Added non-overlapping review comments focused on retry semantics, default behavior for mutating operations, raw helper bypass outside the existing DCR comment, and remaining invalid configuration handling.

internal/cmd/root/products/konnect/common/common.go
harshadixit12: @rspurgeon I need your input for this one - what is the user experience we want?
Should retries be enabled by default?
Should users be able to configure read and write retries separately? I think it could be confusing

rspurgeon: The "ambiguous 5xx" is a good callout for mutating calls. We have a few resources that do not honor unique `name` identifiers for example, so a retry on a `CREATE` could potentially create multiple resources if the backend created the resource but failed in a subsequent way and reported failure.  I also agree that different configurations could be confusing but we could potentially provide both a broad configuration and granular read/write values with the granular overriding the general config. Will continue to consider this...

internal/konnect/httpclient/retry_config.go
harshadixit12: @rspurgeon another item which I wanted to highlight - the SDK retry doesn't support max retries exactly - https://www.speakeasy.com/docs/sdks/customize/runtime/retries#sdk-usage-override, but rather maxElapsedTime

rspurgeon: @harshadixit12 feel free to dismiss invalid review items if needed

internal/konnect/helpers/dcr_providers.go
```diff
@@ -113,9 +115,10 @@ func (a *DCRProvidersAPIImpl) ListDcrProviders(ctx context.Context,
 func (a *DCRProvidersAPIImpl) ListDcrProviderPayloads(
 	ctx context.Context,
 	request kkOPS.ListDcrProvidersRequest,
+	opts ...kkOPS.Option,
 ) (*DCRProviderListPayload, error) {
 	if !a.canUseRawRequest() {
```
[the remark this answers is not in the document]
harshadixit12: We will ignore this one for now, ideally we should use sdk for all calls.

internal/cmd/root/products/konnect/common/common.go
harshadixit12: We fallback to default in this case.
If intentionally a invalid value is passed, it gets validated in the next block.

internal/konnect/httpclient/retry_config.go
```diff
@@ -0,0 +1,56 @@
...
+		Strategy:              rc.Strategy,
+		RetryConnectionErrors: rc.RetryConnectionErrors,
+	}
+	if rc.Strategy == RetryStrategyBackoff {
+		cfg.Backoff = &sdkretry.BackoffStrategy{
+			InitialInterval: rc.InitialIntervalMS,
+			MaxInterval:     rc.MaxIntervalMS,
+			Exponent:        rc.BackoffFactor,
+			MaxElapsedTime:  rc.MaxIntervalMS * rc.MaxAttempts,
+		}
```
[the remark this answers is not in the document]
harshadixit12: The values are capped for both.

internal/declarative/state/client.go
harshadixit12: We are ignoring this for now - ideally we will use sdk for all calls.

internal/cmd/root/products/konnect/declarative/declarative.go
```diff
@@ -1925,7 +1942,11 @@ func runDelete(command *cobra.Command, args []string) error {
 		}
 
 		// Create planner
-		stateClient := createStateClient(kkClient)
+		retryConfig, err := konnectcommon.ResolveRetryConfig(cfg)
+		if err != nil {
+			return err
+		}
+		stateClient := createStateClient(kkClient, []kkOps.Option{retryConfig.ToSDKOption()})
```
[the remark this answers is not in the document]
harshadixit12: Checking this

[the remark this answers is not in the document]
harshadixit12: done in https://github.com/Kong/kongctl/pull/1047/commits/e884dc9b931172e5864b520f12fd043c34c69221

harshadixit12:
Closing this as https://github.com/Kong/kongctl/pull/1072 replaces this with better suitability.

PR #1072, opened 2026-05-15
Feat: implement custom http client with retry

harshadixit12:
Replacement to https://github.com/Kong/kongctl/pull/1047
Implements https://github.com/Kong/kongctl/issues/732

## Summary
Adds support for retries with backoff for **declarative engine** - `plan, diff, apply, sync, and delete` commands.
Since we want retry options to be configurable, we will use a custom HTTP client that supports retry.

Retries are supported on HTTP codes `403, 429, 500, 502, 503, 504`, as well as on connection errors for [safe HTTP methods](https://datatracker.ietf.org/doc/html/rfc7231#section-4.2.2) - `GET, PUT, DELETE`

The HTTP client implementing retry supports exponential backoff, and uses jitter.

The retries are logged with timestamp, and the attempt number.

## Default configuration
```
        http-retry-max-attempts: 3
        http-retry-initial-interval: 1000
        http-retry-max-interval: 60000 #1m 
        http-retry-backoff-factor: 2.0
        http-retry-on-connection-errors: false
```

### Flags
```
--http-retry-max-attempts (int)
--http-retry-initial-interval (int - ms)
--http-retry-max-interval (int - ms)
--http-retry-backoff-factor (float)
--http-retry-on-connection-errors (bool)
```

### Using config file
```
myprofile:
  konnect:
        http-retry-max-attempts: 4
        http-retry-initial-interval: 500
        http-retry-max-interval: 120000 #2 mins
        http-retry-backoff-factor: 1.5
        http-retry-on-connection-errors: true
```

### To disable
```
--http-retry-max-attempts 1 //flag

// config-file
myprofile:
  konnect:
        http-retry-max-attempts: 1
```

## What is not covered
- Respecting `Retry-After` header


internal/cmd/root/verbs/del/del.go
```diff
@@ -105,6 +106,23 @@ Setting this value overrides tokens obtained from the login command.
 	// Add declarative flags from the declarative delete command
 	cmd.Flags().AddFlagSet(declDeleteCmd.Flags())
 
+	// Add retry flags
+	cmd.Flags().Int(cmdcommon.HTTPRetryMaxAttemptsFlagName, 0,
+		fmt.Sprintf(`Maximum total attempts for retryable HTTP requests (0 = use default, 1 disables retries).
+- Config path: [ %s ]`, common.HTTPRetryMaxAttemptsConfigPath))
+	cmd.Flags().Int(cmdcommon.HTTPRetryInitialIntervalFlagName, 0,
```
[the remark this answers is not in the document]
harshadixit12: This is an existing pattern for other flags as well

internal/cmd/root/products/konnect/common/common.go
```diff
@@ -143,6 +150,100 @@ func ResolveHTTPTransportOptions(cfg config.Hook) (httpclient.TransportOptions,
...
+
+	factor, err := resolveOptionalFloat64(cfg, HTTPRetryBackoffFactorConfigPath)
+	if err != nil {
+		return httpclient.RetryConfig{}, err
+	}
+	if factor == 0 {
+		factor = httpclient.DefaultRetryBackoffFactor
+	}
+	if factor < 1 {
+		return httpclient.RetryConfig{}, fmt.Errorf("invalid %s value %g: must be >= 1",
```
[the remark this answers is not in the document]
harshadixit12: fixed in https://github.com/Kong/kongctl/pull/1072/commits/7c18b9ad46a1e05829234e6516cea94b039c44ba

internal/konnect/httpclient/retrying_client.go
```diff
@@ -0,0 +1,325 @@
...
+		resp, err = c.inner.Do(req)
+
+		isLast := attempt == c.cfg.MaxAttempts-1
+
+		if err != nil {
+			if isLast || !c.shouldRetryError(req, err) {
+				return nil, err
+			}
+			next := c.nextInterval(attempt)
+			c.logRetry(req, requestID, 0, attempt+1, next, err.Error())
```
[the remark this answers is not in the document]
harshadixit12: fixed in https://github.com/Kong/kongctl/pull/1072/commits/7c18b9ad46a1e05829234e6516cea94b039c44ba

internal/cmd/root/products/konnect/common/common_test.go
```diff
@@ -211,6 +212,203 @@ func TestResolveHTTPTransportOptions(t *testing.T) {
 	})
 }
 
+func TestResolveRetryConfig(t *testing.T) {
```
harshadixit12, line 218: cc: @rspurgeon I could use another set of eyes on this, and https://github.com/Kong/kongctl/pull/1072/changes#diff-61bbdee5f3ca3bc037f5cc3e13856f8f25716230518c1cb1e2c9cdf789d023e9

I've tried to validate and ensure the values we allow in retry config can't be misused to bombard konnect.

internal/cmd/root/verbs/del/del.go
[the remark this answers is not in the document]
harshadixit12: Will rebase after https://github.com/Kong/kongctl/pull/1074#event-25576642263

internal/konnect/httpclient/retrying_client.go
```diff
@@ -0,0 +1,331 @@
...
+	http.MethodDelete,
+	http.MethodGet,
+	http.MethodHead,
+	http.MethodOptions,
+	http.MethodPut,
+}
+
+// defaultRetryCodes are the HTTP status codes that trigger a retry by default.
+// Matches the SDK's built-in retry status codes.
+var defaultRetryCodes = []int{403, 429, 500, 502, 503, 504}
```
rspurgeon, line 48: This default set does not include `404`, and there is no CLI/config path that feeds `WithRetryableCodes`. The PR goal includes hiding Konnect eventual-consistency `not found` failures, but a GET/read-after-write 404 will currently return immediately. Please either include `404` in the effective retryable set for the declarative retry policy, or expose/configure the status set so the not-found case can actually be enabled.

internal/konnect/httpclient/retrying_client.go
```diff
@@ -0,0 +1,331 @@
...
+		return false
+	}
+
+	method := req.Method
+
+	var urlErr *url.Error
+	if errors.As(err, &urlErr) {
+		// Temporary or timeout errors are safe to retry on any method.
+		if urlErr.Temporary() || urlErr.Timeout() {
+			return true
```
rspurgeon: When `--http-retry-on-connection-errors` is enabled, this retries temporary/timeout URL errors for every method, including POST/PATCH requests with replayable bodies. A client-side timeout does not prove the server failed to process the mutation, so this can duplicate non-idempotent declarative creates/updates. The PR description says connection errors are retried for safe methods; please make this branch use the same idempotent-method gate as EOF/EPIPE/ECONNRESET, or otherwise add an operation-level/idempotency-aware guard.

internal/cmd/root/products/konnect/common/common.go
```diff
@@ -266,6 +401,36 @@ func KonnectSDKFactory(cfg config.Hook, logger *slog.Logger) (helpers.SDKAPI, er
 	}, nil
 }
 
+// This is the real implementation of the SDKAPIFactory,
+// which creates a real Konnect SDK instance
+func KonnectSDKFactory(cfg config.Hook, logger *slog.Logger) (helpers.SDKAPI, error) {
+	retryConfig, err := ResolveRetryConfig(cfg)
```
rspurgeon: `KonnectSDKFactory` now resolves the default retry config unconditionally, so any command path still using `common.GetSDKFactory()` gets retries even if it is not one of the declarative retry verbs. Current examples include direct adopt commands and the view/navigator fallback paths. That bypasses the `GetSDKFactoryForVerb` gating and means 403/429/5xx retries can be enabled by default in command surfaces that do not expose these retry flags. Please keep the ungated factory non-retrying by default, or update all call sites to select retry behavior explicitly.

rspurgeon (changes requested):
I reviewed this as a networking/retry-path change and focused on whether it covers the intended PoLP/eventual-consistency scenarios without widening retry behavior unexpectedly.

The shape is close: body replay is handled before status/error retries, retry config validation has reasonable bounds, and the unit suite plus `make test` pass locally. I found a few issues that should be addressed before merging because they affect the stated behavior and retry safety.

internal/konnect/httpclient/retrying_client.go
rspurgeon: @harshadixit12 I'm not sure if 404 should be covered or not.  I mentioned that as a possibility to the reviewing agent as I'm not sure what return codes we will actually expect to observe in production.

internal/cmd/root/products/konnect/common/common.go
harshadixit12: Valid feedback, will be fixed in next commit.

internal/konnect/httpclient/retrying_client.go
harshadixit12: I think we can skip 404 in the initial implementation. `GET` immediately after `CREATE` is not a common pattern we use.
Also, IMO we should not make HTTP codes for retry configurable through flags - but open to suggestions here.

internal/cmd/root/products/konnect/common/common.go
harshadixit12: done in https://github.com/Kong/kongctl/pull/1072/commits/1034f0670904ec8f8887d5bcac90d96b26703cf9

internal/konnect/httpclient/retrying_client.go
harshadixit12: Done in https://github.com/Kong/kongctl/pull/1072/commits/1034f0670904ec8f8887d5bcac90d96b26703cf9

internal/cmd/root/products/konnect/common/common_test.go
rspurgeon: @harshadixit12 see https://github.com/Kong/kongctl/pull/1072/commits/78876cd61876ccf5020529eb2c308a4ecca58aba

rspurgeon (approved):

PR #1143, opened 2026-05-22
Fix: Support role entity refs by entity type

### LINKED ISSUE #1135: Bug: `teams.roles[].entity_id` is not referenceable
```
organization:
  teams:
    - ref: repro-team
      name: repro-team
      roles:
        - ref: repro-team-role
          role_name: viewer
          entity_id: !ref repro-portal
          entity_type_name: Portals
          entity_region: us

portals:
  - ref: repro-portal
    name: repro-portal
```

```
$ kongctl plan -f repro/team-role-portal-ref.yaml --mode apply
Error: failed to load configuration: organization_team_role "repro-team-role" references portal but expected api: repro-portal (field: entity_id)
```

However, setting the Portal ID without using ref does not result in an error.
From what I can see of the implementation, it seems to reject resource types other than APIs.

rspurgeon:
## Summary

Fixes #1135 by resolving `entity_id` refs for role assignments according to the API-level `entity_type_name` field instead of assuming every role entity ref points at an API.

## What Changed

- Added shared role entity type mapping for `APIs`, `Portals`, and control planes.
- Updated loader validation, resource dependencies, planner dependencies, reference resolution, and executor ref resolution to use that mapping.
- Added unit coverage for portal role entity refs and mismatched role entity refs.
- Added an e2e org team roles scenario covering `entity_id: !ref <portal>#id` with `entity_type_name: Portals`.
- Updated the organization declarative examples to show `!ref` usage in role `entity_id` for API and portal resources.
- Made the e2e binary builder use `-buildvcs=false` and report build output so local worktree VCS stamping failures are actionable.

## Validation

- `go test ./internal/declarative/planner`
- `go test ./internal/declarative/...`
- `go test ./internal/declarative/loader`
- `go test ./internal/declarative/resources`
- `go test -tags=e2e ./test/e2e/harness/...`
- `make test-e2e-scenarios SCENARIO=org/teams/roles`
- `make lint`
- `make test`
- `git diff --check`

`make build` was also verified via `GOFLAGS=-buildvcs=false make build` because this local worktree triggers Go VCS stamping from `/home/rspurgeon` outside the git root.

### DIFF
--- a/docs/examples/declarative/organization/README.md
+++ b/docs/examples/declarative/organization/README.md
@@ -6,9 +6,10 @@ These examples demonstrate how to manage Konnect organization resources with
 ## Files
 
 - `teams.yaml` - defines organization teams with namespaces.
-- `team-roles.yaml` - defines organization teams and assigns roles to an API.
-  The example shows both nested `organization.teams[].roles` declarations and
-  root-level `organization_team_roles` declarations.
+- `team-roles.yaml` - defines organization teams and assigns roles to API and
+  portal resources using `!ref` in `entity_id`. The example shows both nested
+  `organization.teams[].roles` declarations and root-level
+  `organization_team_roles` declarations.
 - `user-assignments.yaml` - assigns existing organization users to teams and
   direct roles. Users have a local `ref` and are selected by exactly one of
   `email` or `id`; team memberships and roles also have local refs. kongctl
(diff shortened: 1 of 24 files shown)


rspurgeon:
cc @imurata

PR #1311, opened 2026-06-13
task(docs): clarify audit-log command workflow

### LINKED ISSUE #708: Docs: Add and improve documentation for kongctl Audit Logs support
There are a variety of commands to enable audit logs. Which one to run when, and with which parameters is unclear. The documentation for flags is also unclear, for example `configure-webhook` defaults to true but how do you disable it?  Can you reuse an existing destination? When to use `tail` vs `listen` vs `ps`

singhvishalkr:
## Problem

The audit-log docs listed the supported command forms, but they did not clearly explain when to use `listen`, `tail`, or the `get audit-logs` inspection commands. They also left some ambiguity around endpoint construction, webhook binding, and whether an existing destination can be reused.

## Root cause

The feature has a few related flows in one page: listener startup, optional streaming, regional webhook binding, and state inspection. Without a short workflow guide, users have to infer behavior from flags and troubleshooting notes.

## Solution

- Add a command selection section for `listen`, `tail`, and `get audit-logs`.
- Clarify `--endpoint` versus `--public-url` + `--path`.
- Document that `listen` creates a session-scoped destination and does not reuse an existing destination.
- Show how to pass `--configure-webhook=false` when webhook configuration is managed separately.

Closes #708

## Verification

- `git diff --check`
- Markdown line-width scan for `docs/audit-logs.md`
- Watermark/trailer scan on the changed docs and commit message

Go tests were not run locally because the Go toolchain is not available on this Windows PATH. The change is documentation-only.

### DIFF
--- a/docs/audit-logs.md
+++ b/docs/audit-logs.md
@@ -14,7 +14,7 @@ including detached process management with `kongctl ps`.
 - Optionally stream events to STDOUT.
 - Optionally run the listener detached in the background.
 
-The feature is exposed through `listen`, `tail`, and `get audit-logs`.
+The feature is exposed through `listen`, `tail`, `get audit-logs`, and `ps`.
 
 ## Command Forms
 
@@ -29,6 +29,9 @@ Supported forms (Konnect-first):
 - `kongctl get audit-logs destinations`
 - `kongctl get audit-logs destination <id|name>`
 - `kongctl get audit-logs webhook`
+- `kongctl ps`
+- `kongctl ps stop <pid>`
+- `kongctl ps stop --all`
 
 Important:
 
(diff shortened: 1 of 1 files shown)


docs/audit-logs.md
```diff
@@ -14,7 +14,7 @@ including detached process management with `kongctl ps`.
 - Optionally stream events to STDOUT.
 - Optionally run the listener detached in the background.
 
-The feature is exposed through `listen` and `tail`.
+The feature is exposed through `listen`, `tail`, and `get audit-logs`.
```
rspurgeon: @singhvishalkr thanks for this, I will review more carefully when I'm back in office next week. I just wanted to note there is also a `ps` command for this feature. I have to review all the commands to be sure this would cover them all to be sure.

singhvishalkr:
Added `ps`, `ps stop <pid>`, and `ps stop --all` to the command overview in `a44266a`, and kept the existing process-registry section below. The audit-log page now lists the detached listener management commands alongside `listen`, `tail`, and `get audit-logs`.

PR #1361, opened 2026-06-23
feat(ai-gateway): [BETA] AI Gateway 2 gateway manager support

### LINKED ISSUE #738: Feat: AI Gateways declarative and get/list support
Parent: #737

Tracks the `AI Gateways` API group from the current AI Gateway OpenAPI spec.

## Scope
- Add declarative configuration support for this resource group in `kongctl`
- Add imperative read support (`get`/`list`) for this resource group
- Wire the resource into `kongctl explain`, `kongctl scaffold`, docs, and tests

## API surface
- `GET /ai-gateways`
- `POST /ai-gateways`
- `GET /ai-gateways/{gatewayId}`
- `PUT /ai-gateways/{gatewayId}`
- `DELETE /ai-gateways/{gatewayId}`
- `GET /ai-gateways/{gatewayId}/expected-config-hash`
- `GET /ai-gateways/{gatewayId}/expected-config-version`

## Acceptance criteria
- Declarative resource support covers the lifecycle exposed by this API group
- Resource matching, planning, and execution handle gateway-scoped relationships where applicable
- Imperative `get` and `list` commands are available for this resource group
- `kongctl explain` and `kongctl scaffold` include this resource
- `docs/declarative-resource-reference.md` and related CLI docs are updated
- Tests cover declarative behavior and imperative read flows

## Notes
- Imperative write commands are out of scope for this issue.
- This is the parent AI Gateway resource; child-resource declarative support will build on its ref and matching model.
- The read-only `expected-config-hash` and `expected-config-version` endpoints are part of this tag; decide whether they need first-pass user-facing command coverage or should remain internal.

### LINKED ISSUE #1472: AI Gateway 2.0: `get` lookup by name is inconsistent — parent matches `display_name` only, child matches `name` only
**Found:** 2026-07-01, AI Gateway 2.0 shakeout against a live `.tech` Konnect org.
**Version:** kongctl `dev` (commit `4bd54436`, build `prerelease-gh-738-aigw-aigw-20260630-0341-g4bd5443`).

## Summary

Every AI Gateway 2.0 entity has two human-facing identifiers: `name` (a slug, e.g. `customer-support-ai-gateway` / `openai-provider` — used throughout the declarative YAML, including for cross-resource references like `provider: openai-provider`) and `display_name` (free text, e.g. `"Customer Support AI Gateway"` / `"OpenAI Provider"`).

`kongctl get` resolves these two identifiers **inconsistently depending on the resource level**, within the very same command family:

- **Parent (`ai_gateway`)** — the top-level positional argument and the `--gateway-name` flag (used by every sub-resource command) resolve **only against `display_name`**. Passing the `name` value fails, and the error message itself says `display_name`, which doesn't help disambiguate what went wrong.
- **Child (e.g. `ai_gateway_provider`)** — the `--provider-name` flag resolves **only against `name`**. Passing the `display_name` value fails.

So there's no single convention a user can rely on — you have to know which resource level you're addressing and use the opposite identifier field depending on level.

## Reproduction

Applied the shipped example (`docs/examples/declarative/ai-gateway/ai-gateway-remote.yaml`), which creates an `ai_gateway` with `name: customer-support-ai-gateway` / `display_name: "Customer Support AI Gateway"`, and a provider with `name: openai-provider` / `display_name: "OpenAI Provider"`.

```bash
# 1. Top-level get by the resource's `name` field — FAILS
kongctl get ai-gateway "customer-support-ai-gateway" -o json
# Error: AI Gateway with display_name "customer-support-ai-gateway" not found

# 2. Top-level get by `display_name` — works
kongctl get ai-gateway "Customer Support AI Gateway" -o json

# 3. Sub-resource lookup via --gateway-name using the resource's `name` field — FAILS
kongctl get ai-gateway providers --gateway-name "customer-support-ai-gateway" -o json
# Error: AI Gateway with display_name "customer-support-ai-gateway" not found

# 4. Sub-resource lookup via --gateway-name using `display_name` — works
kongctl get ai-gateway providers --gateway-name "Customer Support AI Gateway" -o json

# 5. --gateway-id always works reliably, used as workaround
kongctl get ai-gateway providers --gateway-id <id> -o json

# 6. Child-resource --provider-name lookup using the `name` field — works
kongctl get ai-gateway providers --gateway-id <id> --provider-name "openai-provider" -o json

# 7. Same child-resource lookup using `display_name` instead — FAILS (opposite direction from #1/#3)
kongctl get ai-gateway providers --gateway-id <id> --provider-name "OpenAI Provider" -o json
# Error: AI Gateway Provider "OpenAI Provider" not found
```

Re-verified independently on a second gateway state (after other changes had been applied) — same error text, same success/fail split by field both times.

## Impact

- Confusing for anyone scripting right after `apply` — the declarative config and the `kongctl get` docs/examples both treat `name` as the canonical identifier (`kongctl get ai-gateways --jq '.[0].name'` is literally in the shipped usage example), but the parent-level lookup flags/positional args silently require `display_name` instead, while the child-level `--<resource>-name` flags require the opposite (`name`, not `display_name`).
- The error text doesn't disambiguate: the parent-lookup error always says `display_name "..." not found` even when the user supplied what they believed was the `name`, so the fix isn't discoverable from the error alone.
- Affects the top-level `kongctl get ai-gateway <id|name>` command and every `--gateway-name` flag on all 9 sub-resource commands (`agents`, `consumer-groups`, `consumers`, `data-plane-certificates`, `mcp-servers`, `models`, `nodes`, `policies`, `providers`, `vaults`), plus the child-level `--<resource>-name` flags (verified on `--provider-name`).

## Suggested fix

Either (a) resolve name-based lookups against `name` first, falling back to `display_name`, consistently at every resource level — matching how the declarative apply engine resolves cross-resource references (by `name`) — or (b) rename the parent-level flag/error text to `--gateway-display-name` so the required value is unambiguous, and audit whether the same fix is needed for every other `--<resource>-name` flag.

### LINKED ISSUE #1473: Bug: Duplicate `models` entry in `kongctl get ai-gateway --help` command listing
**Found:** 2026-07-01, AI Gateway 2.0 shakeout.
**Version:** kongctl `dev` (commit `4bd54436`, build `prerelease-gh-738-aigw-aigw-20260630-0341-g4bd5443`).

## Reproduction

```bash
kongctl get ai-gateway --help
```

## Observed output (excerpt)

```
Available Commands:
  agents                  List or get Agents for a Konnect AI Gateway
  consumer-groups         List or get Consumer Groups for a Konnect AI Gateway
  consumers               List or get Consumers for a Konnect AI Gateway
  data-plane-certificates List or get data plane certificates for a Konnect AI Gateway
  mcp-servers             List or get MCP Servers for a Konnect AI Gateway
  models                  List or get models for a Konnect AI Gateway
  models                  List or get models for a Konnect AI Gateway
  nodes                   List or get data plane Nodes for a Konnect AI Gateway
  policies                List or get Policies for a Konnect AI Gateway
  providers               List or get providers for a Konnect AI Gateway
  vaults                  List or get Vaults for a Konnect AI Gateway
```

`models` is listed twice, confirmed via `kongctl get ai-gateway --help 2>&1 | grep -n "^  models"` returning two identical lines on two separate runs. Looks like the `models` subcommand is registered twice with Cobra rather than a functional bug — running `kongctl get ai-gateway models --gateway-id <id>` itself works fine and returns correct data with no observed double-execution or duplicate API calls.

## Impact

Cosmetic only, but visible in a prerelease that's about to ship a new command surface — worth a quick fix before GA.

### LINKED ISSUE #1484: AI Gateway 2.0: root ai_gateway's declarative `name:` field is ignored — server stores `ref:` instead
**Found:** 2026-07-01/02, AI Gateway 2.0 POC-criteria implementation against a live `.tech` Konnect org.
**Version:** kongctl `dev` (commit `4bd54436`, build `prerelease-gh-738-aigw-aigw-20260630-0341-g4bd5443`).

## Summary

The root `ai_gateway` resource's declarative `name:` field is silently ignored on create — the server-side `name` ends up set to the fragment's `ref:` value instead. Every child resource (`ai_gateway_provider`, `ai_gateway_policy`, `ai_gateway_consumer`, `ai_gateway_model`, etc.) correctly uses its own explicit `name:` field; only the **root** `ai_gateway` resource exhibits this.

## Reproduction

```yaml
ai_gateways:
  - ref: poc-default-ai-gateway
    name: poc-default-multiprovider
    display_name: "POC Default — Multi-Provider AI Gateway"
    # ...
```

```bash
kongctl apply -f poc-default-ai-gateway-2.0.yaml --auto-approve
kongctl get ai-gateways -o json
```

Observed:

```json
{
  "name": "poc-default-ai-gateway",
  "display_name": "POC Default — Multi-Provider AI Gateway"
}
```

`name` is `poc-default-ai-gateway` — the fragment's `ref:` value — not `poc-default-multiprovider`, the value explicitly declared in `name:`. `display_name` round-trips correctly. Still reproducing live as of 2026-07-02.

## Impact

Low-to-medium — cosmetic today, but it compounds with #1472 (parent-level `kongctl get` lookups resolve against `display_name`, not `name`): the root gateway's `name` field is now doubly unreliable — it's neither settable via declarative config nor used for lookups, even though `name` is the canonical identifier used everywhere else in the declarative model (cross-resource references, child resources' own `name:` fields, etc.).

## Suggested fix

kongctl's create-request builder for `ai_gateway` should send the resource's explicit `name:` field to the API, not fall back to `ref`, when both are present and differ. Worth auditing whether any other root-level (non-nested) AI Gateway 2.0 resource type has the same gap.

### LINKED ISSUE #1486: AI Gateway 2.0: kongctl diff shows perpetual spurious updates for ai_gateway_model / ai_gateway_policy (target order + unmerged server defaults)
**Found:** 2026-07-01/02, AI Gateway 2.0 POC-criteria implementation against a live `.tech` Konnect org.
**Version:** kongctl `dev` (commit `4bd54436`, build `prerelease-gh-738-aigw-aigw-20260630-0341-g4bd5443`).

## Summary

After a clean `apply`, `kongctl diff --mode apply` on the same file still reports `ai_gateway_model` (and, per a broader re-check this session, `ai_gateway_policy` and `ai_gateway_consumer` too) as "to be updated" with zero logical changes — a perpetual spurious diff that defeats `diff`'s core promise (a clean plan should mean nothing to do).

Two causes visible in the `ai_gateway_model` case specifically:

1. **Target order**: the server returns `targets` in whatever order they were created server-side; the diff engine treats array order as significant, so if the locally-declared order doesn't match creation order (easy to happen after any iterative fix-and-reapply), it never resolves to a no-op.
2. **Default expansion**: server-side defaults (e.g. `failover_criteria: [error, timeout]`, `route.protocols: [http, https]`) appear in the remote value but are absent from the diff's "local" comparison value, so the diff never resolves to a true no-op even with identical logical config.

The same underlying pattern (server-side defaults not merged into the "current"/local comparison side before diffing) also produces permanent spurious diffs on `ai_gateway_policy` (extra plugin-schema default fields like `llm_format`, `max_request_body_size`, `genai_category`, `dictionary_name`, `redis: {...}`, etc., appear in the remote value only) — worth treating as the same root cause rather than a model-specific issue.

## Reproduction

```bash
kongctl apply -f poc-default-ai-gateway-2.0.yaml --auto-approve
# ... succeeds, no errors

kongctl diff -f poc-default-ai-gateway-2.0.yaml --mode apply -o text
# Still reports N resources "to be updated" with only server-added default
# fields and/or target-array reordering as the diff — no real logical change.
```

Reproducing live as of 2026-07-02, multiple runs in a row.

## Impact

Medium — this defeats `diff`'s core promise (a clean plan means nothing to do); every `ai_gateway_model` with multiple targets, and apparently every `ai_gateway_policy`, will show a spurious pending change indefinitely, making it hard to tell real drift from noise in CI/GitOps workflows built around `kongctl diff`.

## Suggested fix

- Sort/compare `targets` (and any other unordered nested arrays) by a stable key (e.g. `name`) rather than array position.
- Merge server-side defaults into the "current"/local comparison side of the diff before comparing, matching how other declarative tools (e.g. Terraform) normalize computed/default fields so they don't show as drift.

### LINKED ISSUE #1499: AI Gateway 2.0: declarative `mcp_servers:` block is silently ignored by apply/diff (read-only support only)
**Found:** 2026-07-03, AI Gateway 2.0 POC-criteria implementation against a live `.tech` Konnect org.
**Version:** kongctl `prerelease-aigw-2` (`dev`, commit `832a4eef`, build `2026-07-02T19:56:32Z`).

## Summary

Declaring an `ai_gateway_mcp_server` resource under a top-level `mcp_servers:` key (sibling to `models:`, `policies:`, `providers:`, etc.) in a declarative AI Gateway config file produces **no error, no plan entry, and no create/update action** — it's silently skipped entirely, even though:
- `kongctl get ai-gateway mcp-servers` can read the resource fine once it exists (created out-of-band via the raw Konnect API).
- Other resource types in the exact same file (`policies:`, `models:`, `consumers:`, etc.) are correctly created/updated by `kongctl apply`/`diff`.

This makes `ai_gateway_mcp_server` a read-only resource type from kongctl's declarative-management perspective on this build, despite `get` support existing for it.

## Reproduction

Minimal config, added as a sibling to an existing, working `ai_gateway` block with other resource types already live:

```yaml
ai_gateways:
  - ref: poc-default-ai-gateway
    name: poc-default-multiprovider
    # ... existing policies/models/etc, all working ...
    mcp_servers:
      - ref: poc-mcp-server
        type: passthrough-listener
        name: poc-mcp-server
        display_name: "POC MCP Server (Context7 passthrough)"
        enabled: true
        policies:
          - !ref poc-mcp-oauth2#name
        tools: []
        acl_attribute_type: oauth_access_token
        access_token_claim_field: sub
        config:
          url: "https://mcp.context7.com/mcp"
          route:
            paths: [/mcp]
            methods: [GET, POST]
```

```bash
kongctl diff -f config.yaml
# Plan includes updates for every OTHER resource in the file (policies, models, consumer_groups)
# — zero mention of poc-mcp-server anywhere in the plan output, no error, no warning.

kongctl apply -f config.yaml --auto-approve
# Same — every other resource applies cleanly, poc-mcp-server is never touched.
```

Note: an **earlier**, less-complete version of this same block (missing `acl_attribute_type`/`access_token_claim_field`, which the live API's discriminator actually requires) caused a **different**, harder failure — a client-side YAML/JSON unmarshal error that aborted the *entire file's* parsing before any API calls were made:

```
could not unmarshal `{...}` into expected (Type == passthrough-listener) type
AIGatewayMCPServerPassthroughListener within CreateAIGatewayMCPServerRequest: could not unmarshal
`{...}` into any supported union types for AIGatewayMCPServerPassthroughListener
```

Both symptoms point at the same root gap: MCP server declarative support (create/update via `apply`) doesn't appear to be implemented, whereas `get` (read) support is. Worked around by creating the resource directly via the Konnect API (`POST`/`PUT /v1/ai-gateways/{id}/mcp-servers`) instead — confirmed working end-to-end with real traffic (OAuth2-gated MCP passthrough to a real upstream MCP server) once created that way.

## Impact

Medium-High — MCP is a major AI Gateway 2.0 feature area (multiple `type` modes: `conversion-only`, `conversion-listener`, `listener`, `passthrough-listener`, `upstream-server`), and none of it appears to be declaratively manageable via `kongctl apply` on this build. Anyone trying to manage MCP servers as code (GitOps, CI/CD) can't — they'd need to fall back to raw API calls, defeating the purpose of the declarative tool.

## Suggested fix

Add `ai_gateway_mcp_server` (all 5 `type` variants) to the declarative apply/diff resource-type registry, including proper discriminated-union parsing for the type-specific required fields (`acl_attribute_type`/`access_token_claim_field` for OAuth-ACL variants, `tools_cache_ttl_seconds` for `upstream-server` mode, etc.) matching what the live API's `CreateAIGatewayMCPServerRequest` schema actually requires.

Env: Konnect `.tech`, Org ID `5b8cee88-069e-4840-8a24-384bc634ab9b`, Control Plane/AI Gateway ID `7ef00efe-932a-4127-98a1-22248e05e68f`.

rspurgeon:
## Summary
- Add complete AI Gateway 2 resource support across declarative workflows, imperative get/list commands, and `kongctl view`.
- Cover supported AI Gateway resources with examples and lifecycle e2e scenarios running against the Konnect `.tech` environment.
- Stabilize planning and diff behavior for API-expanded defaults, unordered collections, local-only `ref` values, and same-plan references.
- Validate declarative AI Gateway payloads against current SDK request types and fail when a non-empty field would otherwise be silently discarded.
- Align `kongctl explain` and scaffold schemas with current AI Gateway union variants, including model-provider authentication headers.
- Strengthen MCP server validation and coverage for variant-specific access fields and permissive tool objects such as `query`.
- Build against internal AI Gateway SDK commit `837069df006da0f22a591c71788003e323db304d`.

## Testing
- `go fix ./...`
- `make format`
- `make build`
- `make test-all`
- `make test-e2e-scenarios SCENARIO=ai-gateway/mcp-server` passed against `.tech`; artifacts: `/home/rspurgeon/go/e2e-artifacts/20260713-205040`
- `make test-e2e-scenarios`: all AI Gateway and combined scenarios passed; the run had one unrelated `dump/organization-teams` failure because the configured E2E user was absent from the target org; artifacts: `/home/rspurgeon/go/e2e-artifacts/20260713-221437`

Closes #738
Closes #1473
Closes #1484
Closes #1486
Closes #1499


internal/declarative/planner/ai_gateway_planner.go
```diff
@@ -0,0 +1,364 @@
...
+	byID := make(map[string]state.AIGateway)
+	byDisplayName := make(map[string][]state.AIGateway)
+	for _, gateway := range gateways {
+		if gateway.ID != "" {
+			byID[gateway.ID] = gateway
+		}
+		byDisplayName[gateway.DisplayName] = append(byDisplayName[gateway.DisplayName], gateway)
+	}
+	return byID, byDisplayName
+}
```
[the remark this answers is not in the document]
rspurgeon: Fixed in `7645c26a` by adding a `Name` index and matching managed AI Gateways by the stable Konnect API name/ref before falling back to `display_name`.

internal/declarative/planner/ai_gateway_planner.go
```diff
@@ -0,0 +1,364 @@
...
+		return state.AIGateway{}, false, nil
+	case 1:
+		return matches[0], true, nil
+	default:
+		return state.AIGateway{}, false, fmt.Errorf(
+			"multiple managed AI Gateways with display_name %q found in namespace; use a UUID ref or remove duplicates",
+			desired.DisplayName,
+		)
+	}
+}
```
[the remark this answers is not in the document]
rspurgeon: Fixed in `7645c26a`: `matchCurrentAIGateway` now prefers the desired ref/name for matching and only falls back to `display_name` when the API name is empty, so display-name edits plan as updates.

rspurgeon:
Closes #1473 
Closed #1472

rspurgeon:
Pushed d4f738fb with the latest AI Gateway diff and e2e scenario fixes.

Included:
- consumer-group membership reconciliation and diff handling
- declared-key config comparison for policy and identity-provider opaque config maps
- AI Gateway explain schema fix for nested agents
- e2e scenario cleanup for agent, consumer-group, data-plane-certificate, identity-provider, and vault
- quieter reset behavior for configured E2E users missing from an org
 
Local validation:
- make format
- go test ./internal/declarative/planner ./internal/declarative/resources ./internal/declarative/state ./internal/declarative/executor ./test/e2e/harness -tags=e2e\n- make lint
- make build
 
Note: poc-default-ai-gateway-2.0.yaml remains untracked locally and was not committed.

internal/declarative/planner/ai_gateway_vault_planner.go
```diff
@@ -0,0 +1,302 @@
...
+				desiredVault,
+				updateFields,
+				changedFields,
+				nil,
+				plan,
+			)
+		}
+	}
+
+	if plan.Metadata.Mode == PlanModeSync {
```
rspurgeon: **Bug: protection check skipped on sync deletes — 9 of 11 child planners.**

This sync-delete loop removes unmatched vaults without consulting `kongctl.protected`, while the provider and identity-provider planners in this same PR guard the identical loop:

```go
isProtected := labels.IsProtectedResource(current.NormalizedLabels)
if err := p.validateProtection(ResourceTypeAIGatewayProvider, name, isProtected, ActionDelete); err != nil {
    return err
}
```
(`ai_gateway_provider_planner.go:94-97`, `ai_gateway_identity_provider_planner.go`)

The same unguarded loop exists in the model, policy, agent, consumer, consumer-group, consumer-credential, MCP-server, and data-plane-certificate planners (`grep -L validateProtection internal/declarative/planner/ai_gateway_*_planner.go`). The state types for these children all populate `NormalizedLabels`, and `p.validateProtection` is generic, so the fix is mechanical — but as written, a resource labeled protected is silently deleted during `sync`. None of the `ai_gateway_*_planner_test.go` files exercise protection, which is likely how this slipped through; a test on at least one child type would lock it in.

internal/declarative/planner/dependencies.go
```diff
@@ -224,6 +231,45 @@ func (d *DependencyResolver) ResolveDependenciesWithGroups(
...
+		ResourceTypeAIGatewayConsumerGroup,
+		ResourceTypeAIGatewayModel,
+		ResourceTypeAIGatewayMCPServer,
+		ResourceTypeAIGatewayVault,
+		ResourceTypeAIGatewayDataPlaneCertificate:
+	default:
+		return ""
+	}
+
+	if change.Parent != nil {
```
rspurgeon, line 257: **Bug: consumer-credential changes escape per-gateway serialization.**

`ResourceTypeAIGatewayConsumerCredential` is listed in the switch above, but this Parent-first lookup defeats it: every other child planner sets `Parent` to the gateway, while the credential planner sets `Parent: &ParentInfo{Ref: consumerRef, ID: consumerID}` (`ai_gateway_consumer_credential_planner.go:239`). Credentials therefore key off `"ref:" + consumerRef` and never reach the `References[FieldAIGatewayID]` fallback.

Reproduced with a minimal scenario (one gateway, consumers X/Y, credential under X): `ResolveDependenciesWithGroups` places `create consumerY` and `create credX` in the same execution group, and `executor.go` dispatches each group concurrently via `errgroup` — so a credential mutation can race other children of the same gateway.

Fix options: consult `References[FieldAIGatewayID]` before `Parent` in this function, or have the credential planner carry the gateway in `References` and check that first. `TestResolveDependenciesWithGroups_AIGatewayChildrenSameParentSerialized` doesn't cover credentials — worth extending.

internal/cmd/root/products/konnect/adopt/ai_gateway.go
```diff
@@ -0,0 +1,204 @@
...
+		if resp == nil {
+			break
+		}
+		list := resp.ListAIGatewaysResponse
+		if list == nil || len(list.Data) == 0 {
+			break
+		}
+
+		for _, gateway := range list.Data {
+			if gateway.DisplayName == identifier {
```
rspurgeon: **Bug: `adopt` resolves by DisplayName only; `get` resolves by Name first.**

Every sibling adopt implementation matches on `.Name` (`event_gateway.go:200`, `api.go:174`, `dashboard.go:188`, `portal.go:214`), and `getAIGateway.go` was explicitly fixed in this PR (2d9f02dd) to check `Name` before falling back to `DisplayName`. This loop only checks `DisplayName`, so `kongctl adopt konnect ai-gateway <name>` returns "not found" for any gateway whose Name and DisplayName differ, even though `kongctl get ai-gateway <name>` resolves it fine.

Suggest mirroring the name-first / display-name-fallback logic from `getAIGateway.go`. `ai_gateway_test.go` currently only covers DisplayName lookup, so a Name-lookup case would close the gap.

internal/cmd/root/verbs/dump/common.go
```diff
@@ -158,6 +158,32 @@ func mapResourceName(name string) string {
...
+		"ai_gateway_model_provider", "ai_gateway_model_providers":
+		return "ai_gateway_model_providers"
+	case "ai-gateway-identity-provider", "ai-gateway-identity-providers",
+		"ai_gateway_identity_provider", "ai_gateway_identity_providers":
+		return "ai_gateway_identity_providers"
+	case "ai-gateway-policy", "ai-gateway-policies", "ai_gateway_policy", "ai_gateway_policies":
+		return "ai_gateway_policies"
+	case "ai-gateway-agent", "ai-gateway-agents", "ai_gateway_agent", "ai_gateway_agents":
+		return "ai_gateway_agents"
+	case "ai-gateway-consumer", "ai-gateway-consumers", "ai_gateway_consumer", "ai_gateway_consumers":
```
rspurgeon, line 173: **Bug: missing hyphenated alias for consumer credentials.**

Every other AI Gateway child type in this switch gets a hyphen+underscore alias pair, but there is no `case "ai-gateway-consumer-credential", "ai-gateway-consumer-credentials", ...` arm. Since `default:` returns the input unchanged, `--resources ai-gateway-consumer-credentials` fails `validateResourceList` (only `ai_gateway_consumer_credentials` is in `declarativeAllowedResources`), while every sibling accepts both forms. Verified: `mapResourceName("ai-gateway-consumer-credentials")` returns the input unmapped. The new alias tests don't exercise this type either.

internal/declarative/planner/ai_gateway_provider_planner.go
```diff
@@ -0,0 +1,325 @@
...
+	if provider.Labels != nil {
+		fields[FieldLabels] = provider.Labels
+	}
+	if provider.ManagedBy != nil {
+		fields[FieldManagedBy] = provider.ManagedBy
+	}
+	return fields
+}
+
+func aiGatewayProviderConfigChanged(current, desired map[string]any) bool {
```
rspurgeon, line 317: **Pattern: provider config comparison bypasses the shared normalization pipeline — perma-diff risk.**

`aiGatewayProviderConfigChanged` JSON-round-trips and scrubs secrets, then does a strict `reflect.DeepEqual`. Every sibling comparator (model, policy, agent, consumer, consumer-group, mcp-server, vault) runs `normalizeAIGatewayPayloadsForComparison` first, which prunes nils, server-expanded defaults, and empty containers (`ai_gateway_payload_compare.go`) — the machinery this PR built precisely for idempotent re-plans.

Because `current.Config` comes from the raw API response while `desired.Config` is only what the user declared, any server-returned null or optional default sub-field in provider auth config produces a spurious update on every `plan` after `apply`. The vault planner shows the intended composition (shared pipeline, then secret scrubbing layered on top). No test covers provider re-plan idempotency.

internal/declarative/loader/loader.go
```diff
@@ -1225,6 +1306,79 @@ func (l *Loader) extractNestedResources(rs *resources.ResourceSet) {
 		}
 		egw.DataPlaneCertificates = nil
 	}
+
+	for i := range rs.AIGateways {
```
rspurgeon: **Pattern: second extraction loop over `rs.AIGateways` duplicates the first.**

This loop re-extracts IdentityProviders, Policies, Agents, Consumers, ConsumerGroups, Models, MCPServers, and Vaults — all already extracted and nilled by the earlier loop at ~1039-1096. Only `Providers` and `DataPlaneCertificates` are genuinely new here. It works today only because the fields are already nil, but it becomes a real double-extraction bug (duplicate entries reaching planner/executor) if the loops ever drift or reorder, with no compiler or lint signal. The Event Gateway extraction just above (~1270-1308) shows the house pattern: one pass per family. Suggest folding `Providers` and `DataPlaneCertificates` into the first loop and deleting this one.

internal/cmd/root/verbs/dump/declarative.go
```diff
@@ -656,6 +840,523 @@ func collectDeclarativeEventGateways(
...
+	gateways, err := collectDeclarativeAIGateways(ctx, aiGatewayClient, requestPageSize, filterOptions{})
+	if err != nil {
+		return nil, err
+	}
+
+	var providers []declresources.AIGatewayProviderResource
+	for _, gateway := range gateways {
+		gatewayProviders, err := buildAIGatewayProviders(
+			ctx,
+			slog.Default(),
```
rspurgeon: **Pattern: `slog.Default()` instead of the configured logger.**

`collectDeclarativeAIGatewayProviders` (here) and `collectDeclarativeAIGatewayIdentityProviders` (~line 1002) pass `slog.Default()` into `buildAIGatewayProviders` / `buildAIGatewayIdentityProviders`, which emit `logWarn(...)` for missing name/type/display_name/config. Every other collector/populate function in this file threads the logger obtained via `helper.GetLogger()`, so warnings on these two root-level dump paths bypass `--log-level` and the configured handler.

internal/declarative/planner/ai_gateway_model_planner.go
```diff
@@ -0,0 +1,391 @@
...
+		return nil
+	}
+
+	var deps []string
+	for _, targetModel := range targetModels {
+		target, ok := targetModel.(map[string]any)
+		if !ok {
+			continue
+		}
+		providerName, ok := target["provider"].(string)
```
rspurgeon: **Pattern:** `FieldProvider` (`constants.go:190`) already exists for this key and is used for the same semantic field in `ai_gateway_payload_compare.go:124` — per the repo rule against new string literals for plan fields, this should be `target[FieldProvider]`.

internal/cmd/root/products/konnect/aigateway/policies.go
```diff
@@ -0,0 +1,451 @@
...
+		records,
+		policies,
+		"",
+		tableview.WithCustomTable(
+			[]string{
+				aiGatewayHeaderID,
+				aiGatewayHeaderName,
+				aiGatewayHeaderDisplayName,
+				aiGatewayHeaderType,
+				"ENABLED",
```
rspurgeon: **Pattern (consolidated): `"ENABLED"` literal where the package constant exists.**

`aiGatewayHeaderEnabled` (`child_common.go:104`) is used for this column by `agents.go` (both views), `models.go:477`, and `mcp_servers.go:427`, but the raw literal appears here and at `policies.go:432`, `models.go:263`, and `mcp_servers.go:246` — each file inconsistent even with its own child-view builder. Worth converging on the constant since the package already established it (adjacent `"GLOBAL"` could get the same treatment if kept).

internal/cmd/root/products/konnect/aigateway/getAIGateway.go
```diff
@@ -0,0 +1,441 @@
...
+	LocalCreatedTime string
+	LocalUpdatedTime string
+}
+
+type getAIGatewayCmd struct {
+	*cobra.Command
+}
+
+func aiGatewayToDisplayRecord(gateway kkComps.AIGateway) aiGatewayDisplayRecord {
+	const missing = "n/a"
```
rspurgeon: Minor: this file declares a local `const missing = "n/a"` here and at line 343, while all ~11 sibling command files in the package use the shared `aiGatewayMissingValue` (`child_common.go:87`). Since this file is the template the siblings were stamped from, a future sentinel change would silently miss it.

rspurgeon (commented):
## Review: AI Gateway resource support

Structured review of the full 290-file change set, sharded by architectural layer (cross-cutting security sweep, loader/validator, planner, executor/state, resources, SDK helpers/dump/adopt, command surface, e2e fixtures, docs). Each shard was reviewed against the house pattern distilled from `docs/contributor/declarative-resource-implementation-guide.md` and the event-gateway family, and every candidate finding was independently re-verified against the code (several with throwaway reproduction tests) before posting. 19 candidates were raised; 13 survived verification; 6 were refuted and dropped.

### Findings (inline comments, severity-ordered)

**Likely bugs (4)**
1. Sync-mode deletes skip the `kongctl.protected` check in 9 of the 11 AI Gateway child planners (comment on `ai_gateway_vault_planner.go`).
2. Consumer-credential changes escape the per-gateway execution serialization the code explicitly opts them into (`dependencies.go`) — reproduced empirically.
3. `adopt ai-gateway` resolves by DisplayName only, unlike `get ai-gateway` and every sibling adopt command (`adopt/ai_gateway.go`).
4. `dump --resources` has no hyphenated alias for consumer credentials, the only child type missing one (`dump/common.go`).

**Pattern conformance (5)**
- Provider config comparison bypasses the shared `normalizeAIGatewayPayloadsForComparison` pipeline — perma-diff risk on re-plan (`ai_gateway_provider_planner.go`).
- Duplicated child-extraction loop over `rs.AIGateways` in the loader — latent double-processing hazard (`loader.go`).
- Two dump collectors log via `slog.Default()` instead of the configured logger (`dump/declarative.go`).
- `"provider"` literal where `FieldProvider` exists (`ai_gateway_model_planner.go`).
- `"ENABLED"` header literal in 3 files where siblings use `aiGatewayHeaderEnabled` (consolidated comment on `policies.go`).

**Minor (1)**: local `missing` constant duplicating `aiGatewayMissingValue` in `getAIGateway.go`.

### Verified clean — security focus

The security sweep specifically targeted the secret-bearing resource types this PR introduces, and found the handling sound:

- Consumer credential `api_key` is stripped from payload maps, rejected by `Validate()`, and excluded from detail views and dump reconstruction.
- Vault and identity-provider secrets (tokens, `client_secret`) cannot leak via get/list/dump: the SDK response types are secret-less `...Output` variants, and the identity-provider planner additionally scrubs `client_secret` from plan/diff `ChangedFields` — stricter than the existing portal IdP precedent.
- `RedactSensitiveFields` is wired into executor debug logging with clone-then-redact semantics, covered by a new unit test.
- CI workflow changes route all PR/label/dispatch-controlled values through `env:` blocks rather than `${{ }}` interpolation inside `run:` scripts, avoiding script injection; secret handling relies on GitHub's automatic masking as before.
- No hardcoded secrets in e2e fixtures or examples; data-plane-certificate fixtures are public self-signed certs with no private keys.

One adjacent polish note from a refuted finding: root-level empty `ai_gateway_model_providers:` / `ai_gateway_identity_providers:` lists **are** safely blocked in sync mode by the generic `validateSyncScope` guard (so no scope-widening risk), but they are the only 2 of the 11 child collections without the friendly type-specific loader error in `sync_scope.go`'s `captureRootChildScope` — worth aligning while in there.

---
*Review method: layer-sharded multi-agent review with independent adversarial verification of each finding, coordinated via Claude Code.*

rspurgeon (commented):
## Review: AI Gateway resource family

Reviewed with a layered multi-agent process: the diff was sharded by architectural layer (resources / loader / planner / executor+state / SDK helpers / commands / e2e / docs, plus a cross-cutting security sweep), each shard reviewed against the house pattern distilled from `docs/contributor/declarative-resource-implementation-guide.md` and the merged event-gateway family, and every candidate finding was independently adversarially verified before posting. 30 raw findings were reduced to the 14 inline comments below (2 security, 5 correctness, 7 pattern conformance).

### Security

Two inline findings, both in the planner secret handling:

1. **Vault secret rotation never reaches the API** — the vault planner returns the *scrubbed* payload as the update fields, so write-only secrets are stripped from `UpdateAIGatewayVaultRequest` (`ai_gateway_vault_planner.go:224`). This is the highest-priority finding.
2. **Cleartext provider secrets in plan/diff output** — the provider planner stores raw config in `ChangedFields`, and diff redaction only keys on top-level field names (`ai_gateway_provider_planner.go:248`). The identity-provider planner already scrubs at the equivalent site, so this is a one-file fix.

**Checked and found clean** (so silence here is coverage, not omission):

- HTTP trace logging redacts bearer tokens and secret-bearing body fields for all AI Gateway SDK traffic (`internal/konnect/httpclient`); executor debug logging wraps `change.Fields` in `RedactSensitiveFields`.
- Consumer credential `api_key` is stripped from every payload path, rejected in `Validate()`, and the consumer e2e scenario asserts `api_key == null` on re-read — credentials are correctly create/delete-only.
- SDK response (`*Output`) types omit vault/provider/identity-provider secrets on read paths, so `get`/`list`/`dump` cannot echo stored secrets.
- `.github/workflows` changes keep the `pull_request` trigger, fork detection, and environment-gated secrets; no new exposure of `GH_TOKEN_PRIVATE_READ` or Konnect PATs to fork code.
- E2E fixtures and examples use obviously-fake placeholder values.

**Acknowledged, not flagged**: secret-only rotations being undetectable for write-only fields is documented behavior (`docs/declarative.md` "Write-only Secret Fields") with precedent in the portal IdP planner — the provider/identity-provider compare-side scrubbing follows that contract. The vault finding above is flagged because it goes further and drops the secret from the update payload itself, which the same doc says must not happen.

### Correctness

- Sync-mode child pruning has no external-gateway guard, unlike event-gateway — destructive for adopted gateways (`ai_gateway_vault_planner.go:86`, applies family-wide).
- Empty `ai_gateway_providers:` / `ai_gateway_identity_providers:` root collections silently widen sync delete scope instead of erroring like the nine guarded siblings (`sync_scope.go:538`).
- Provider/identity-provider planners match current state by name only — renames produce duplicates instead of updates (`ai_gateway_provider_planner.go:47`).
- `getParentType` omits agents, weakening plan dependency ordering (`dependencies.go:350`).
- Agent child view drops a column its list view and siblings have (`agents.go:420`).

### Pattern conformance

Inline comments cover: bare-name `GetKonnectMonikerFilter` (family-wide), `slog.Default()`/`nil` loggers in dump collectors, duplicated extraction loops in `extractNestedResources`, policy child-view column order, hardcoded table-header literals, `DisplayRecord` naming divergence, and a raw `"ai_gateway"` schema literal.

### Minor notes (no inline comment)

- `test/e2e/scenarios/explain/command-coverage/scenario.yaml` has two steps numbered `014-` (lines 285/311, names differ) — renumber for tidiness.
- `docs/examples/declarative/ai-gateway/ai-gateway-remote.yaml:27` uses an inline placeholder for an auth header value; consider demonstrating `!env` there, since that field is exactly where users will paste real secrets.
- Executor and state layers were reviewed in full; no findings.

internal/declarative/planner/ai_gateway_vault_planner.go
```diff
@@ -0,0 +1,302 @@
...
+	if err != nil {
+		return false, nil, nil, fmt.Errorf("failed to normalize desired AI Gateway Vault %q: %w", desired.Ref, err)
+	}
+
+	currentCompare, desiredCompare := normalizeAIGatewayPayloadsForComparison(currentPayload, desiredPayload)
+	currentCompare = scrubAIGatewayVaultWriteOnlyFields(currentCompare).(map[string]any)
+	desiredCompare = scrubAIGatewayVaultWriteOnlyFields(desiredCompare).(map[string]any)
+
+	currentPlanPayload := scrubAIGatewayVaultWriteOnlyFields(currentPayload).(map[string]any)
+	desiredPlanPayload := scrubAIGatewayVaultWriteOnlyFields(desiredPayload).(map[string]any)
```
rspurgeon, line 229: **[security]** `shouldUpdateAIGatewayVault` scrubs the write-only secret fields (`token`, `secret_access_key`, `client_secret`, `key`, `secret_id`, `api_key`) out of `desiredPlanPayload` and then returns that same scrubbed map as the update fields (line 231). `AIGatewayVaultAdapter.MapUpdateFields` (`internal/declarative/executor/ai_gateway_vault_adapter.go:51`) marshals `change.Fields` directly into `UpdateAIGatewayVaultRequest`, so a declared secret value is never sent on update: rotating a HashiCorp token or AWS/Azure/Conjur secret alongside another field change reports success while silently omitting the new secret from the wire request. This breaks the documented write-only contract in `docs/declarative.md` ("The initial create or update still sends the configured secret value").

The sibling `ai_gateway_provider_planner.go` gets this right: scrubbing is confined to `aiGatewayProviderConfigChanged` for comparison, and `updateFields = extractAIGatewayProviderFields(desired)` uses the raw config. Suggest the vault planner do the same — scrub only for compare, return the unscrubbed desired payload as update fields. (`TestAIGatewayVaultPlannerIgnoresWriteOnlySecretDrift` currently locks in the scrubbed-payload behavior and would need updating.)

internal/declarative/planner/ai_gateway_provider_planner.go
```diff
@@ -0,0 +1,325 @@
...
+			New: labels.GetUserLabels(desired.Labels),
+		}
+	}
+
+	if desired.ManagedBy != nil && !reflect.DeepEqual(current.ManagedBy, desired.ManagedBy) {
+		changedFields[FieldManagedBy] = FieldChange{Old: current.ManagedBy, New: desired.ManagedBy}
+	}
+
+	if desired.Config != nil && aiGatewayProviderConfigChanged(current.Config, desired.Config) {
+		changedFields[FieldConfig] = FieldChange{Old: current.Config, New: desired.Config}
```
rspurgeon: **[security]** This `FieldChange` stores the raw, unscrubbed configs — including write-only secrets (`config.auth.headers[].value`, `client_secret`, `secret_access_key`, `service_account_json`) — into `PlannedChange.ChangedFields`, which is serialized into `plan -o json` output and rendered by the human diff. Redaction in the diff renderer (`isSensitiveDiffField`, `internal/cmd/root/products/konnect/declarative/declarative.go:1589`) keys on the top-level field name only, and `config` is not sensitive, so changing any non-secret provider field prints the cleartext secret into plan files and diff output.

The identity-provider planner already handles this correctly at its equivalent site (`ai_gateway_identity_provider_planner.go:255` wraps Old/New in `scrubAIGatewayIdentityProviderSecretFields`), and `scrubAIGatewayProviderSecretFields` already exists in this file — apply it to the Old/New values here.

internal/declarative/planner/ai_gateway_vault_planner.go
rspurgeon: **[bug]** This sync-mode prune loop (and its equivalents in the other AI Gateway child planners) deletes any server-side child not declared locally, with no external-gateway guard. `planExternalAIGatewayChildren` (`ai_gateway_planner.go:386`) routes external gateways into this same function, so declaring a single vault under an adopted/external `ai_gateway` and running `sync` will plan deletion of every other vault on that gateway. The event-gateway exemplar guards exactly this case: `if plan.Metadata.Mode == PlanModeSync && !p.isEventGatewayExternal(gatewayRef)` (`internal/declarative/planner/event_gateway_backend_cluster_planner.go:131`). There is no `isAIGatewayExternal` helper in this family — one is needed, applied to every child prune loop.

internal/declarative/loader/sync_scope.go
```diff
@@ -398,15 +528,45 @@ func captureSyncScope(content []byte, rs *resources.ResourceSet) error {
...
-func captureRootChildScope(scope *resources.SyncScope, raw map[string]any, entry childCollectionScope) {
+func captureRootChildScope(scope *resources.SyncScope, raw map[string]any, entry childCollectionScope) error {
 	value, ok := raw[entry.key]
 	if !ok {
-		return
+		return nil
 	}
 	items, ok := asSlice(value)
 	if !ok || len(items) == 0 {
+		if entry.resourceType == resources.ResourceTypeAIGatewayPolicy {
```
rspurgeon, line 547: **[bug]** Nine of the eleven AI Gateway child collection types get an explicit "cannot be empty" error in this block, but `ResourceTypeAIGatewayProvider` and `ResourceTypeAIGatewayIdentityProvider` are missing. For those two, an empty root collection (`ai_gateway_providers: []` / `ai_gateway_identity_providers: []`) falls through to `scope.AddRootChildCollection(...)`, which marks the whole resource type as a global delete-scope candidate instead of failing like every sibling — a partial sync file can silently put all model providers / identity providers in deletion scope. The loader tests mirror the gap: `TestLoaderRejectsRootLevelEmpty...` exists for the nine guarded types but not these two.

internal/declarative/planner/ai_gateway_provider_planner.go
```diff
@@ -0,0 +1,325 @@
...
+		p.planAIGatewayProviderCreatesForNewGateway(namespace, gatewayRef, gatewayName, gatewayChangeID, desired, plan)
+		return nil
+	}
+
+	currentProviders, err := p.client.ListAIGatewayProviders(ctx, gatewayID)
+	if err != nil {
+		return fmt.Errorf("failed to list AI Gateway Model Providers for gateway %s: %w", gatewayID, err)
+	}
+
+	currentByName := make(map[string]state.AIGatewayProvider)
```
rspurgeon: **[bug]** `planAIGatewayProviderChanges` builds only a `currentByName` map and matches desired providers strictly by `Name`. Every other AI Gateway child planner (agents, vaults, models, consumers, consumer groups, credentials, MCP servers, policies) builds both `currentByID` and `currentByName` and checks ID first (see `matchCurrentAIGatewayAgent`, `internal/declarative/planner/ai_gateway_agent_planner.go:284`). A provider renamed in Konnect (or matched via an adopted UUID ref) misses the lookup and lands in the create branch, producing a duplicate instead of an update. `planAIGatewayIdentityProviderChanges` has the identical gap.

internal/declarative/planner/dependencies.go
```diff
@@ -301,6 +347,18 @@ func (d *DependencyResolver) getParentType(childType string) string {
 		return ResourceTypeAPI
 	case ResourceTypePortalPage:
 		return ResourceTypePortal
+	case ResourceTypeAIGatewayProvider,
```
rspurgeon, line 370: **[pattern]** `getParentType` lists nine AI Gateway child types here but omits `ResourceTypeAIGatewayAgent`, which is a recognized child type elsewhere in this file (line 239). `findParentChange` therefore cannot resolve an agent's parent gateway change in either `ResolveDependencies` path, so agent changes can lose parent dependency ordering in generated plans.

internal/declarative/resources/ai_gateway.go
```diff
@@ -0,0 +1,441 @@
...
+	for i := range a.Vaults {
+		a.Vaults[i].SetDefaults()
+	}
+	for i := range a.DataPlaneCertificates {
+		a.DataPlaneCertificates[i].SetDefaults()
+	}
+}
+
+// GetKonnectMonikerFilter returns the filter string for Konnect API lookup.
+func (a AIGatewayResource) GetKonnectMonikerFilter() string {
```
rspurgeon, line 319: **[pattern]** `GetKonnectMonikerFilter` returns the bare name; the established contract (`BaseResource.GetKonnectMonikerFilter`, `internal/declarative/resources/base.go:63`, used by portal and event-gateway types) is a `name[eq]=<name>` filter expression. Most of the AI Gateway family shares this deviation. If nothing consumes the filter for these types today it is latent, but generic code that builds list queries from the moniker filter will silently mis-filter. Suggest `return a.BaseResource.GetKonnectMonikerFilter(a.Name)` for consistency (with the event-gateway-style `// TODO: the API does not support filtering by name` comment if applicable).

internal/cmd/root/verbs/dump/declarative.go
rspurgeon: **[pattern]** The standalone AI Gateway provider/identity-provider collectors use `slog.Default()` (here and line 1002), and `collectDeclarativeAIGatewayDataPlaneCertificates` passes a literal `nil` logger (line 1335), while every other collector threads the configured logger from `helper.GetLogger()` (line 200). With `slog.Default()`, dump warnings bypass kongctl's configured log level and format; with `nil`, `logWarn` silently drops the missing-ID/missing-content warnings these builders emit. All three call sites should receive the command's logger.

internal/declarative/loader/loader.go
rspurgeon: **[pattern]** `extractNestedResources` contains two near-identical `for i := range rs.AIGateways` loops (line 1039 and here) extracting overlapping child collections. The first loop nils each field after extraction, so the duplicated extractions in the second loop are dead code — but the duplication is a trap: a future child type added to one loop and not the other will silently behave differently depending on which copy runs first. The two loops should be merged.

internal/cmd/root/products/konnect/aigateway/agents.go
```diff
@@ -0,0 +1,449 @@
...
+	tableRows := make([]table.Row, 0, len(agents))
+	for i := range agents {
+		record := aiGatewayAgentToRecord(agents[i])
+		tableRows = append(tableRows, table.Row{
+			record.ID,
+			record.Name,
+			record.DisplayName,
+			record.Type,
+			record.Enabled,
+			record.PolicyCount,
```
rspurgeon, line 420: **[pattern]** `buildAIGatewayAgentChildView` omits the `LocalUpdatedTime` column that `listAgents` includes (line 229) and that sibling child views include (e.g. `buildAIGatewayModelChildView` in `models.go`). The nested/child navigation shows fewer columns than the top-level list for the same resource type.

internal/cmd/root/products/konnect/aigateway/policies.go
```diff
@@ -0,0 +1,451 @@
...
+		fmt.Fprintf(&b, "%s: %s\n", field, formatAIGatewayModelDetailValue(payload[field]))
+	}
+	return strings.TrimRight(b.String(), "\n")
+}
+
+func buildAIGatewayPolicyChildView(policies []kkComps.AIGatewayPolicy) tableview.ChildView {
+	tableRows := make([]table.Row, 0, len(policies))
+	for i := range policies {
+		record := aiGatewayPolicyToRecord(policies[i])
+		tableRows = append(tableRows, table.Row{
```
rspurgeon, line 416: **[pattern]** Column order here is `Type` before `DisplayName`, while `listPolicies` (lines 223-229) renders `DisplayName` before `Type`; sibling child views match their list-view order. The headers below also diverge from the row order fix once columns are aligned.

internal/cmd/root/products/konnect/aigateway/models.go
```diff
@@ -0,0 +1,496 @@
...
+		records,
+		models,
+		"",
+		tableview.WithCustomTable(
+			[]string{
+				aiGatewayHeaderID,
+				aiGatewayHeaderName,
+				aiGatewayHeaderDisplayName,
+				aiGatewayHeaderType,
+				"ENABLED",
```
rspurgeon: **[pattern]** Literal `"ENABLED"` header instead of the `aiGatewayHeaderEnabled` constant (`child_common.go:104`) used elsewhere. The same literal appears in `mcp_servers.go:246`, `policies.go:248`, and `policies.go:432` (which also hardcodes `"GLOBAL"` — worth adding a constant alongside the fix). Worth one sweep so the header constants stay the single source of truth.

internal/cmd/root/products/konnect/aigateway/providers.go
```diff
@@ -0,0 +1,540 @@
...
+%[1]s get ai-gateways model-providers --gateway-name "Customer Support Gateway"
+# Get a model provider by ID or name
+%[1]s get ai-gateways model-providers --gateway-name "Customer Support Gateway" openai-provider
+# Get a model provider by flag
+%[1]s get ai-gateways model-providers --gateway-id <gateway-id> --model-provider-name openai-provider
+`, meta.CLIName)),
+	)
+)
+
+type aiGatewayProviderDisplayRecord struct {
```
rspurgeon: **[pattern]** `aiGatewayProviderDisplayRecord` (and `aiGatewayIdentityProviderDisplayRecord`, `identity_providers.go:53`) break the `aiGateway<Type>Record` naming used by the other ten record types in this package. Renaming keeps the copy-paste family greppable.

internal/declarative/resources/ai_gateway_provider.go
```diff
@@ -0,0 +1,231 @@
...
+	a.Labels = raw.Labels
+	a.ManagedBy = raw.ManagedBy
+	a.Config = raw.Config
+	return nil
+}
+
+func aiGatewayProviderExplainNode(_ ExplainBuildContext) (*ExplainNode, error) {
+	return explainObject(
+		explainResourceRefField(),
+		explainRefField("ai_gateway", ResourceTypeAIGateway, true),
```
rspurgeon: **[pattern]** Raw `"ai_gateway"` literal in the explain schema where this file already uses `SchemaFieldAIGateway` (lines 130/160/172) and sibling types pass the constant to `explainRefField` (e.g. `ai_gateway_agent.go:367`, `ai_gateway_consumer.go:354`). AGENTS.md forbids new string literals for contract identifiers like child parent-ref fields.

rspurgeon:
Pushed `da0c25de` addressing the rspurgeon AI Gateway review findings.

Covered in this commit:
- sync delete external-parent guards across AI Gateway child planners
- protection checks on child sync deletes where state/API responses expose labels
- provider and identity-provider ID-first matching before name fallback
- vault update payloads preserve write-only values while plan/diff output stays scrubbed
- provider config diff scrubbing and shared AI Gateway comparison normalization
- consumer credential execution serialization by AI Gateway
- adopt/dump aliases, loader duplicate extraction cleanup, schema/field constants, TUI table nits, and regression coverage

Two notes from the review summary:
- AI Gateway data-plane certificate responses in the current SDK (`AIGatewayDataPlaneClientCertificate`) do not include labels, so there is no reliable `kongctl.protected` source for that delete path yet. I added the external-gateway sync guard there, but did not invent protection state client-side.
- I left `docs/examples/declarative/ai-gateway/ai-gateway-remote.yaml` self-contained rather than switching its provider header to `!env`; the local `ai-gateway.yaml` already demonstrates `!env OPENAI_AUTH_HEADER`, while the remote example is meant to work via `-f https://...` without local tag/env prerequisites.

Local validation before push:
- `make format`
- `make lint`
- `make test`
- `make build`
- `make test-e2e-scenarios SCENARIO=explain/command-coverage`

rspurgeon:
Addressed the latest #1486 follow-up drift report in one batch.

Changes pushed in d0bf1822:
- Normalize AI Gateway policy references before comparing, including resolved __REF__ values and server-returned ordering, so agent policy lists do not produce no-op updates.
- Treat default MCP access values (acl_attribute_type=consumer, empty acls/default_tool_acls) as ignorable when the API reads conversion-only MCP servers back with access omitted/null.
- Added planner coverage for both cases and expanded the MCP server e2e scenario to declare the default access shape.

Validated locally:
- go fix ./...
- make format\n- make lint
- make build\n- make test
- go test ./internal/declarative/planner

rspurgeon:
Updated the internal AI Gateway SDK replace to rjs-aigw commit 3acb2fa8 via 3a827531.\n\nValidated locally after the update:\n- make build\n- make lint\n- make test\n\nThe previous SDK compile failure around missing AIGatewayIdentityProvider*Response types is resolved by this SDK generation.

rspurgeon:
Updated model-provider authentication handling in commit 03b34950.

- `explain` and `scaffold` now emit the API-supported `config.auth.headers[].name/value` shape.
- Deprecated flat `header_name`/`header_value` input now fails validation instead of being silently discarded by SDK decoding.
- Added unit coverage and expanded the model-provider e2e scenario to verify header-name persistence and write-only value omission.
- Documented model-provider write-only credential behavior.

Validated with `go fix ./...`, `make format`, `make build`, `make lint`, and `make test`. The live tech e2e scenario remains for CI/manual execution because this sandbox cannot read the local PAT file.

PR #1537, opened 2026-07-08
fix(konnect): report unknown profile instead of loading an empty config

### LINKED ISSUE #1363: Bug: invalid profile configuration names are not detected
In this example there is no profile in the loaded configuration named `tech`

```
./kongctl --profile tech get org
Error: error sending request: refresh Konnect access token after HTTP 401: refresh Konnect access token: failed to refresh token: 401 Unauthorized
```

The tool should detect that the user is trying to load an invalid configuratin profile and provide a userful error instead of loading an empty configuration

apoorva-01:
Pass `--profile` a name that doesn't exist and kongctl says nothing. It loads an empty config, then later dies on a vague "authentication token not available" that never mentions the profile. A typo shouldn't send you debugging your creds.

Now it checks first: if the profile isn't one it knows (default, in your config, `KONGCTL_<PROFILE>_*` env vars, or a token from a past login), it just says `profile "tech" is not configured`. Test covers the unknown case and all four known ones.

One call I'd want your take on: I put the check on the auth path, not config load. Config load also runs for `login`/`--help`/completion, so checking there would break `login --profile new` on first setup. Trade-off is it only fires on auth commands. Easy to move if you'd rather.

Closes #1363

### DIFF
--- a/internal/cmd/root/products/konnect/common/common.go
+++ b/internal/cmd/root/products/konnect/common/common.go
@@ -9,6 +9,7 @@ import (
 	"net/url"
 	"os"
 	"regexp"
+	"slices"
 	"strconv"
 	"strings"
 	"time"
@@ -540,11 +541,7 @@ func accessTokenUnavailableError(cfg config.Hook) error {
 	profile := cfg.GetProfile()
 
 	if !profileIsConfigured(cfg) {
-		return fmt.Errorf(
-			"profile %q is not configured. Run '%s get profiles' to list configured "+
-				"profiles, or '%s login --profile %s' to set it up",
-			profile, meta.CLIName, meta.CLIName, profile,
-		)
+		return unknownProfileError(cfg, profile)
 	}
 
 	envVar := fmt.Sprintf("KONGCTL_%s_KONNECT_PAT", strings.ToUpper(profile))
@@ -573,7 +570,7 @@ func profileIsConfigured(cfg config.Hook) bool {
 	if name == kprofile.DefaultProfile {
 		return true
 	}
-	if profileInConfigFile(cfg, name) {
+	if slices.Contains(configuredProfileNames(cfg), name) {
 		return true
 	}
 	if profileHasEnvVars(name) {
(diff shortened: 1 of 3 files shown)


rspurgeon:
Thanks @apoorva-01, very helpful.

Some minor thoughts. As of now with `kongctl` there is no function to "setup a profile", the config file is ready-only from the CLI perspective. What happens with `kongctl login --profile <name>` is that a token is created with the given profile name, so authentication going forward would work.

Could we tweak this error message to avoid saying that `login --profile <name>` will “set up” a profile? It would also be helpful if we listed the profiles for the user. There is a function available to grab them.

I’d suggest wording like this:

```
profile "tech" is not configured.

Available profiles:
    default
    dev

Use one of these profiles or run 'kongctl login --profile tech' to authenticate with this profile.
```

If no profiles are available, maybe:

```
profile "tech" is not configured.

No profiles are configured yet.

Run 'kongctl login --profile tech' to authenticate with this profile.
```

The newline-separated profile list should make typo recovery easier when the user simply picked the wrong profile name.

Thanks again for your contribution!

apoorva-01:
Thanks, the pointer about `login` not actually setting up a profile is the useful bit here. I'd assumed it created the profile rather than just a token for that name, so the old wording was misleading. Fixed that.

It now lists the configured profiles too, and falls back to "No profiles are configured yet." when there aren't any. Went with your suggested wording pretty much as-is. Pushed.

internal/cmd/root/products/konnect/common/common.go
```diff
@@ -553,6 +561,68 @@ func accessTokenUnavailableError(cfg config.Hook) error {
...
+	var b strings.Builder
+	fmt.Fprintf(&b, "profile %q is not configured.\n\n", profile)
+
+	available := configuredProfileNames(cfg)
+	if len(available) > 0 {
+		b.WriteString("Available profiles:\n")
+		for _, name := range available {
+			fmt.Fprintf(&b, "    %s\n", name)
+		}
+		fmt.Fprintf(&b, "\nUse one of these profiles or run '%s login --profile %s' "+
```
rspurgeon: Security concern: `profile` is user-controlled and this renders it inside a shell-shaped command without escaping. Values with quotes, spaces, or shell metacharacters can make the guidance misleading or unsafe to copy.

I recommend avoiding direct command guidance here and rendering the profile as data instead, for example: `Run kongctl login with --profile set to %q to authenticate with this profile.` The same safer wording should be used in the no-profiles branch below.

apoorva-01: Both branches now render the profile with %q instead of splicing it into the command, so a name with quotes or spaces shows up as escaped data rather than something you can paste and run. Added a test with a metacharacter profile name to lock it in.

internal/konnect/auth/token_source.go
```diff
@@ -188,6 +189,13 @@ func credentialFilePath(cfg config.Hook) string {
 	return filepath.Join(cfgPath, getCredentialFileName(profile))
 }
 
+// HasStoredCredential reports whether a saved login credential file exists on
+// disk for the configured profile.
+func HasStoredCredential(cfg config.Hook) bool {
+	_, err := os.Stat(credentialFilePath(cfg))
+	return err == nil
+}
```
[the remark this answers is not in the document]
apoorva-01: Updated. `HasStoredCredential` now only treats `os.ErrNotExist` as absent, so a permission or transient IO error counts as present instead of flipping the profile to "unknown".

rspurgeon:
Thank you @apoorva-01 i'll work through final testing and merge. Appreciate the contributions!

rspurgeon:
@apoorva-01 closing and reopening this PR in order to pickup changes to our CI flows

rspurgeon:
Reopening to retrigger updated fork-safe E2E status checks after #1570.

rspurgeon (approved):

Issue #1532, opened 2026-07-07
Fix: Fix misleading API version spec scaffold and resource docs


rspurgeon:
## Summary

A user hit a failed `kongctl apply --plan` when uploading an API spec through
declarative config because our helper output and generated resource docs point
them toward a misleading API version spec shape:

```yaml
versions:
  - ref: petstore-v1
    version: "1.0.27"
    spec:
      content: !file ../../../oas.yaml
```

That input currently plans successfully, but the saved plan contains a
double-wrapped spec payload. The API version change ends up with
`fields.spec.content` set to a JSON string whose top-level key is `content`,
instead of the OpenAPI document itself:

```json
{
  "content": {
    "openapi": "3.0.4",
    "info": {},
    "paths": {}
  }
}
```

Konnect then rejects the API version create with:

```text
content must be valid specification
```

The working/canonical shape is:

```yaml
versions:
  - ref: petstore-v1
    version: "1.0.27"
    spec: !file ../../../oas.yaml
```

## Why this is misleading

The incorrect shape is easy to infer from our own helper commands and resource
docs:

- `kongctl scaffold api.versions` currently emits:

  ```yaml
  apis:
    - ref: my-resource
      versions:
        - version: v1.0.0
          spec:
            # content: !file ./specs/openapi.yaml
          ref: my-resource
  ```

- `kongctl scaffold api_version` emits the same `spec.content` pattern.
- `kongctl explain api.versions` and `kongctl explain api_version` describe
  `spec` as an object with a `content` property and mark both `spec` and
  `content` with `x-kongctl-preferred-tag: !file`.
- `docs/declarative-resource-reference.md` documents:

  ```yaml
  spec: object required
    content: string required (OpenAPI or AsyncAPI content; json or yaml) # prefer: !file ./specs/api.yaml
  ```

- That generated resource reference is used by the public docs page, which now
  has the same misleading guidance:
  https://developer.konghq.com/kongctl/supported-resources/#apis

Some narrative examples already use the better shape (`spec: !file ...`), so
the guidance is inconsistent.

## Expected behavior

Users should be guided to the canonical config:

```yaml
spec: !file ./specs/openapi.yaml
```

`spec.content: !file ./specs/openapi.yaml` should not successfully produce a
plan when the loaded file is structured YAML/JSON. That input shape currently
creates a valid-looking plan artifact that only fails later during `apply`,
after Konnect rejects the malformed spec payload. The configuration loader or
[...]

PR #1567, opened 2026-07-09
fix(declarative): scaffold api version spec as a file scalar

apoorva-01:
`kongctl scaffold`/`explain` for API versions point you at the wrong spec shape:

```
spec:
  # content: !file ./specs/openapi.yaml
```

That double-wraps the payload (`{"content": {...openapi...}}`), which plans fine but fails apply with `content must be valid specification`. The shape that works is `spec: !file ./specs/api.yaml`.

So this shows `spec` as a `!file` scalar in scaffold and explain, and fixes the same thing in the resource-reference doc.

I left out the loader validation from the checklist on purpose. Rejecting a `spec.content: !file` config before planning is a separate design call (reject vs normalize vs warn) and touches the planner, so better as a follow-up once you've picked the behavior. Happy to take it.

Part of #1532.

### DIFF
--- a/docs/declarative-resource-reference.md
+++ b/docs/declarative-resource-reference.md
@@ -111,7 +111,7 @@ apis:
     versions: # https://developer.konghq.com/api/konnect/api-builder/v3/#/operations/create-api-version
       - ref: string
         version: string
-        spec: object required (OpenAPI or AsyncAPI content; json or yaml) # prefer: !file ./specs/api.yaml
+        spec: !file ./specs/api.yaml # required (OpenAPI or AsyncAPI content; json or yaml)
     publications: # https://developer.konghq.com/api/konnect/api-builder/v3/#/operations/publish-api-to-portal
       - ref: string
         portal_id: string required (uuid) # prefer: !ref <portal-ref>
--- a/docs/declarative-resource-reference.md
+++ b/docs/declarative-resource-reference.md
@@ -111,8 +111,7 @@ apis:
     versions: # https://developer.konghq.com/api/konnect/api-builder/v3/#/operations/create-api-version
       - ref: string
         version: string
-        spec: object required
-          content: string required (OpenAPI or AsyncAPI content; json or yaml) # prefer: !file ./specs/api.yaml
+        spec: object required (OpenAPI or AsyncAPI content; json or yaml) # prefer: !file ./specs/api.yaml
     publications: # https://developer.konghq.com/api/konnect/api-builder/v3/#/operations/publish-api-to-portal
       - ref: string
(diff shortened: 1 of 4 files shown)


internal/declarative/resources/api_version.go
```diff
@@ -13,7 +13,20 @@ func init() {
 	registerResourceType(
 		ResourceTypeAPIVersion,
 		func(rs *ResourceSet) *[]APIVersionResource { return &rs.APIVersions },
-		AutoExplain[APIVersionResource](),
+		AutoExplain[APIVersionResource](
+			// spec is an SDK object ({content: ...}) but is supplied whole from a
+			// file. Present it as `spec: !file ...` so users don't nest the content
+			// under spec.content, which double-wraps the payload and Konnect rejects.
+			WithExplainFieldHint("spec", ExplainFieldHint{
+				Literal:      "!file ./specs/openapi.yaml",
```
rspurgeon: Small consistency nit: the docs and PR description use `./specs/api.yaml`, while the scaffold hint uses `./specs/openapi.yaml`. Either works, but keeping the sample path consistent across scaffold, explain, and docs would make the guidance easier to follow.

```suggestion
				Literal:      "!file ./specs/api.yaml",
```

docs/declarative-resource-reference.md
```diff
@@ -111,8 +111,7 @@ apis:
     versions: # https://developer.konghq.com/api/konnect/api-builder/v3/#/operations/create-api-version
       - ref: string
         version: string
-        spec: object required
-          content: string required (OpenAPI or AsyncAPI content; json or yaml) # prefer: !file ./specs/api.yaml
+        spec: object required (OpenAPI or AsyncAPI content; json or yaml) # prefer: !file ./specs/api.yaml
```
rspurgeon: This is the shape users should actually write, so it may be clearer to show the tagged scalar directly instead of describing `spec` as an object. The underlying SDK field is object-shaped, but the declarative guidance is trying to steer users away from `spec.content`.

```suggestion
        spec: !file ./specs/api.yaml # required (OpenAPI or AsyncAPI content; json or yaml)
```

internal/declarative/resources/api_version.go
apoorva-01: Everything uses `./specs/api.yaml` now (scaffold hint, explain, docs), so no more mixed sample paths.

docs/declarative-resource-reference.md
apoorva-01: Agreed, that reads better. Took your suggestion as-is.

rspurgeon (approved):

PR #1575, opened 2026-07-09
fix(konnect): report all missing token create flags at once

### LINKED ISSUE #1155: Task: Bundle command flag and configuration errors
If you take the following example commands, first the user is told they need `--name`, and after providing that they need to provide an expire flag. Both flags are required. These errors could be bundled and help the user in one shot. 

```
  └─❯❯ ./kongctl create pat
  Error: --name is required

  Run 'kongctl create pat --help' for usage

  └─❯❯ ./kongctl create pat --name "rick-del-me"
  Error: exactly one of --expires-in or --expires-at is required

  Run 'kongctl create pat --help' for usage
```

One example could be:
```
└─❯❯ ./kongctl create pat
  Error: --name is required
  Error: --expires-in or --expires-at is required

  Run 'kongctl create pat --help' for usage
```

There are multiple software patterns to choose from to solve this issue, the implementor is free to choose the best pattern for the kongctl codebase.

apoorva-01:
`create pat` checks required flags one at a time, so you rerun it just to find the next missing one. Now it collects them with `errors.Join` and reports all at once (same for `create spat`).

The list rendering lives in the shared usage-error path (`renderCommandUsageError`), so any command that returns an `errors.Join` wrapped in a `ConfigurationError` gets one `- Error:` line per failure. Single errors render exactly as before. `ConfigurationError` now has `Unwrap`, so the joined set stays inspectable.

Closes #1155.

### DIFF
--- a/internal/cmd/helper.go
+++ b/internal/cmd/helper.go
@@ -175,6 +175,12 @@ func (e *ConfigurationError) Error() string {
 	return e.Err.Error()
 }
 
+// Unwrap exposes the wrapped error so callers (and the top-level error
+// renderer) can inspect a joined set of validation errors.
+func (e *ConfigurationError) Unwrap() error {
+	return e.Err
+}
+
 func (e *ExecutionError) Error() string {
 	return e.Err.Error()
 }
--- a/internal/cmd/root/products/konnect/token/token.go
+++ b/internal/cmd/root/products/konnect/token/token.go
@@ -1,6 +1,7 @@
 package token
 
 import (
+	"errors"
 	"fmt"
 	"reflect"
 	"strconv"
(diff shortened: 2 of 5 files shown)


rspurgeon:
Thanks again @apoorva-01!

I'd be interested in improving the handling of errors and specifically error lists globally to create a better pattern.  Maybe we should make the global usage-error renderer understand joined validation errors instead of accepting the single `Error:`

I think this can stay as `errors.Join`, but we should make the global usage-error renderer understand joined validation errors instead of accepting the single `Error:` header.

Right now commands return one `error`, and `root.Execute` eventually calls `renderCommandUsageError`, which prints `Error: %s` using `err.Error()`. Since `errors.Join` renders as newline-separated text, the second validation message loses the `Error:` prefix.

Could you move this behavior into the shared error path instead?

One possible shape:
- add `Unwrap() error` to `cmd.ConfigurationError`, so wrapped joined errors remain inspectable
- add a small helper near `renderCommandUsageError` that flattens errors implementing `Unwrap() []error`
- have `renderCommandUsageError` print each flattened validation error as `- Error: <message>`
- keep single errors rendering exactly as they do today
- add root-level tests for `kongctl create pat` / `kongctl create spat` missing required flags

That gives commands a reusable pattern: collect validation failures with `errors.Join`, wrap once in `ConfigurationError`, and let the top-level renderer decide how to display a set of usage errors.

apoorva-01:
Done, pushed. Moved it into the shared path like you suggested: `Unwrap()` on `ConfigurationError`, a small flatten helper, and each bundled failure prints as `- Error: ...` (single errors unchanged). Added root-level tests for `create pat`/`create spat` too.

PR #1639, opened 2026-07-16
fix(cmd): list resource paths when explain or scaffold run with no args

### LINKED ISSUE #1629: Task: `kongctl explain` and `kongctl scaffold` executed bare (no args) must display available arguments
These should show what resources are available to explain/scaffold

```
kongctl explain
```

```
kongctl scaffold
```

apoorva-01:
`kongctl explain` and `kongctl scaffold` with no arguments just fail with `accepts 1 arg(s), received 0`, which doesn't tell you what you can actually pass. Now they print the available resource paths.

The list comes from the explain registry rather than a hand-maintained one, so it can't drift. `organization` and `analytics` are left out on purpose, they only resolve with a child segment like `analytics.dashboards`, so listing them bare would advertise a path that errors.

One existing test needed updating: `TestRootErrorUX` used bare `scaffold` to check that argument errors show a concise help hint. That case isn't an error anymore, so it now uses `scaffold api portal` instead, which still exercises the same hint.

Worth a look: with no args it prints the plain list regardless of `--output`. Happy to make it honour json/yaml if you'd rather.

Closes #1629.


rspurgeon (changes requested):
Thank you so much @apoorva-01 .  I think i'd prefer we show the nested path by default for all the resource types.  Can we shift the guidance to be like this:
```
portal
portal.page
portal.snippet
```
etc...

rspurgeon:
@apoorva-01 I went ahead and made this update directly on the branch. Bare `explain` and `scaffold` now prefer valid nested resource paths such as `portal.pages`, `portal.snippets`, and `api.versions`, while retaining canonical root paths for resources without a supported nested form. I also updated the coverage and ran the unit and relevant E2E scenario tests. Thanks again for the contribution!

rspurgeon:
@apoorva-01 I pushed the organization assignment follow-up directly to your branch. The explain/scaffold paths are now organization.users.teams, organization.users.roles, organization.system-accounts.teams, and organization.system-accounts.roles, while the four existing flat aliases continue to resolve. I also added proper root-level declarative support for the four assignment resource keys, including selector validation and sync-scope handling. Existing nested declarative input remains supported unchanged; this is additive to the accepted input format.

PR #1704, opened 2026-07-25
fix(declarative): resolve parent_policy_id ref for modify_headers and skip_record policies

### LINKED ISSUE #1606: `parent_policy_id: !ref` causes HTTP 500 on produce/consume policies
### Title

`kongctl apply` fails with HTTP 500 when produce/consume policies use `parent_policy_id: !ref`

### Labels

`bug`, `event-gateway`

### Body

**Description**

When applying an Event Gateway configuration that includes produce or consume policies with `parent_policy_id: !ref <name>`, `kongctl apply` resolves the `!ref` tag correctly but the resulting API call returns an HTTP 500 Internal Server Error from Konnect. The same policies created via the REST API (passing the resolved ID as a plain string `parent_policy_id`) succeed without issues.

**Environment**

- kongctl: latest (Homebrew, `brew install --cask kong/kongctl/kongctl`)
- macOS: arm64 (Apple Silicon)
- Konnect region: US (`us.api.konghq.com`)
- Date tested: 2026-07-14

**Input configuration**

```yaml
event_gateways:
  - ref: event-gateway-quickstart
    name: event-gateway-quickstart

    backend_clusters:
      - ref: retail-kafka-cluster
        name: retail-kafka-cluster
        bootstrap_servers:
          - kafka1:9092
        authentication:
          type: anonymous
        insecure_allow_anonymous_virtual_cluster_auth: true
        tls:
          enabled: false

    virtual_clusters:
      - ref: storefront
        name: storefront
        destination:
          name: retail-kafka-cluster
        dns_label: storefront
        acl_mode: enforce_on_gateway
        authentication:
          - type: anonymous
        produce_policies:
          # This parent policy applies successfully
          - ref: validate-orders-produce
            type: schema_validation
            name: validate-orders-produce
            config:
              type: json
              value_validation_action: reject
          # This child policy triggers HTTP 500
          - ref: flag-high-value-orders
            type: modify_headers
            name: flag-high-value-orders
            parent_policy_id: !ref validate-orders-produce
            condition: "record.value.content.amount > 500"
            config:
              actions:
                - op: set
                  key: x-priority
                  value: high
        consume_policies:
          - ref: validate-orders-consume
            type: schema_validation
            name: validate-orders-consume
            config:
              type: json
              value_validation_action: mark
          # This child policy also triggers HTTP 500
          - ref: filter-test-orders
            type: skip_record
            name: filter-test-orders
            parent_policy_id: !ref validate-orders-consume
            condition: "record.value.content.status == \"test-order\""
```

**Steps to reproduce**

```bash
# 1. Create the Event Gateway and backend cluster first (these succeed)
kongctl apply -f foundations.yaml

# 2. Apply the file with child policies
kongctl apply -f transformations.yaml
```

**Expected behavior**

All policies, including child policies with `parent_policy_id: !ref <parent-name>`, are created successfully. The `!ref` tag should resolve to the parent policy's ID and be passed as a plain string in the API call.

**Actual behavior**

The parent policies (`validate-orders-produce`, `validate-orders-consume`) are created successfully. The child policies (`flag-high-value-orders`, `filter-test-orders`) fail with HTTP 500 Internal Server Error.

**Workaround**

Create parent policies with `kongctl apply`, then create child policies via the Konnect REST API:

```bash
# Get the parent policy ID
PRODUCE_SV_ID=$(curl -s "$KONNECT_API/v1/event-gateways/$EG_ID/virtual-clusters/$VC_ID/produce-policies" \
  -H "Authorization: Bearer $KONNECT_TOKEN" | jq -r '.[] | select(.name=="validate-orders-produce") | .id')

# Create child policy with plain string parent_policy_id
curl -X POST "$KONNECT_API/v1/event-gateways/$EG_ID/virtual-clusters/$VC_ID/produce-policies" \
  --fail-with-body -H "Authorization: Bearer $KONNECT_TOKEN" \
  --json '{"type": "modify_headers", "name": "flag-high-value-orders",
           "parent_policy_id": "'$PRODUCE_SV_ID'",
           "condition": "record.value.content.amount > 500",
           "config": {"actions": [{"op": "set", "key": "x-priority", "value": "high"}]}}'
```

**Affected policy types (all require `parent_policy_id`)**

- `modify_headers` (produce phase)
- `skip_record` (consume phase)
- `encrypt_fields` (produce phase)
- `decrypt_fields` (consume phase)

apoorva-01:
`modify_headers` (produce) and `skip_record` (consume) child policies with `parent_policy_id: !ref` ship the raw `__REF__:...` placeholder to Konnect, which 500s.

The planner's parent-policy helpers only pulled the ID off the `encrypt_fields`/`decrypt_fields` variants, so `prepareProducePolicyParentRefs`/`prepareConsumePolicyParentRefs` skipped those two types and never swapped the placeholder for the resolved parent ID. Both variants carry `parent_policy_id` in the SDK, so this just teaches the helpers to read and write it the same way the encrypt/decrypt ones already do.

Unit tests on both `prepare*ParentRefs` cover the `modify_headers`/`skip_record` case with the parent already in current state. They fail on main (placeholder leaks through) and pass with the fix.

closes #1606

### DIFF
--- a/internal/declarative/planner/event_gateway_consume_policy_planner.go
+++ b/internal/declarative/planner/event_gateway_consume_policy_planner.go
@@ -575,22 +575,33 @@ func consumePolicyIsSchemaValidation(policy resources.EventGatewayConsumePolicyR
 }
 
 func consumePolicyParentPolicyID(policy resources.EventGatewayConsumePolicyResource) string {
-	if policy.EventGatewayParsedRecordDecryptFieldsPolicyCreate == nil {
+	switch {
+	case policy.EventGatewaySkipRecordPolicyCreate != nil:
+		if id := policy.EventGatewaySkipRecordPolicyCreate.ParentPolicyID; id != nil {
+			return *id
+		}
+		return ""
+	case policy.EventGatewayParsedRecordDecryptFieldsPolicyCreate != nil:
+		return policy.EventGatewayParsedRecordDecryptFieldsPolicyCreate.ParentPolicyID
+	default:
 		return ""
 	}
-	return policy.EventGatewayParsedRecordDecryptFieldsPolicyCreate.ParentPolicyID
 }
 
 func consumePolicyWithParentPolicyID(
 	policy resources.EventGatewayConsumePolicyResource,
 	parentPolicyID string,
 ) resources.EventGatewayConsumePolicyResource {
-	if policy.EventGatewayParsedRecordDecryptFieldsPolicyCreate == nil {
-		return policy
+	switch {
+	case policy.EventGatewaySkipRecordPolicyCreate != nil:
+		variant := *policy.EventGatewaySkipRecordPolicyCreate
+		variant.ParentPolicyID = &parentPolicyID
+		policy.EventGatewaySkipRecordPolicyCreate = &variant
+	case policy.EventGatewayParsedRecordDecryptFieldsPolicyCreate != nil:
+		variant := *policy.EventGatewayParsedRecordDecryptFieldsPolicyCreate
+		variant.ParentPolicyID = parentPolicyID
+		policy.EventGatewayParsedRecordDecryptFieldsPolicyCreate = &variant
 	}
-	variant := *policy.EventGatewayParsedRecordDecryptFieldsPolicyCreate
-	variant.ParentPolicyID = parentPolicyID
-	policy.EventGatewayParsedRecordDecryptFieldsPolicyCreate = &variant
 	return policy
 }
 
--- a/internal/declarative/planner/event_gateway_consume_policy_planner.go
+++ b/internal/declarative/planner/event_gateway_consume_policy_planner.go
@@ -576,6 +576,11 @@ func consumePolicyIsSchemaValidation(policy resources.EventGatewayConsumePolicyR
 
 func consumePolicyParentPolicyID(policy resources.EventGatewayConsumePolicyResource) string {
 	switch {
+	case policy.EventGatewayModifyHeadersPolicyCreate != nil:
+		if id := policy.EventGatewayModifyHeadersPolicyCreate.ParentPolicyID; id != nil {
+			return *id
+		}
+		return ""
 	case policy.EventGatewaySkipRecordPolicyCreate != nil:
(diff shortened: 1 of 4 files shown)


internal/declarative/planner/event_gateway_consume_policy_planner.go
```diff
@@ -575,22 +575,33 @@ func consumePolicyIsSchemaValidation(policy resources.EventGatewayConsumePolicyR
 }
 
 func consumePolicyParentPolicyID(policy resources.EventGatewayConsumePolicyResource) string {
-	if policy.EventGatewayParsedRecordDecryptFieldsPolicyCreate == nil {
+	switch {
+	case policy.EventGatewaySkipRecordPolicyCreate != nil:
```
rspurgeon, line 584: Could we handle `EventGatewayModifyHeadersPolicyCreate` in both consume helpers here?

The SDK exposes `modify_headers` as a consume-policy variant, and its shared create schema has an optional `parent_policy_id`. A valid consume `modify_headers` child referencing an existing `schema_validation` parent will still return an empty ID here, so `prepareConsumePolicyParentRefs` skips resolution and the raw `__REF__` placeholder can reach Konnect—the same failure this PR fixes for the other variants.

Please add the corresponding read/write cases and a consume `modify_headers` regression test.

rspurgeon (changes requested):
Requesting changes for the remaining consume-policy variant that supports `parent_policy_id`.

internal/declarative/planner/event_gateway_consume_policy_planner.go
rspurgeon: Addressed in 4421ae09: consume `modify_headers` policies now read and replace `parent_policy_id` references, with a focused regression test. Unit, integration, lint, and the consume-policy E2E scenario pass.

rspurgeon (approved):
The requested consume modify_headers parent-reference support is implemented in 4421ae09 and verified locally.

rspurgeon:
Working on unblocking CI w/ unrelated errors off main

PR #1826, opened 2026-08-07
feat(declarative): support explicit writes for write-only secrets

### LINKED ISSUE #1748: feat(declarative): support explicit writes for write-only secrets
## Summary

Add first-class secret-write intents to kongctl's declarative planning and
execution flow.

Konnect has fields that accept secret material on create or update but do not
return the value from GET or list operations. Because kongctl is stateless, it
cannot compare the desired value with the current remote value. The planner
currently omits these fields from comparison, which avoids perpetual updates
but also means a secret-only change does not produce a plan.

Users need an explicit, reviewable way to say "write this configured secret
during this execution." The plan must record that intent without containing
the secret value.

## Core model

Secret handling has two independent concerns:

1. **Write intent:** whether this plan should write a write-only field.
2. **Value source:** where the executor obtains the value.

A deferred source such as !env answers the second question, but it must not
implicitly answer the first. Normal plans must remain idempotent unless the
user explicitly requests a secret write.

For an unreadable remote value, kongctl can only:

1. write it every time;
2. require an explicit write request; or
3. store a comparable version or fingerprint somewhere.

Kongctl should use the second option. It preserves the stateless model without
creating perpetual updates or inventing a comparison against an unknowable
remote value.

## Proposed user experience

Support write selection at three scopes:

~~~sh
# Write one secret field.
kongctl plan -f konnect.yaml \
  --write-secret workforce-idp#config.client_secret \
  --output-file rotation.json

# Write every configured write-only secret on one resource.
kongctl plan -f konnect.yaml \
  --write-secret workforce-idp \
  --output-file rotation.json

# Write every supported, configured write-only secret in the input.
kongctl plan -f konnect.yaml \
  --write-secrets \
  --output-file rotation.json
~~~

The exact selector grammar may reuse an existing kongctl resource selector if
one is suitable. A short resource ref should work when unambiguous. Ambiguous
refs must produce an error with the fully qualified alternatives.

The flags are authoring shortcuts for first-class plan intents. They must not
act as generic switches that make planners pretend every secret differs.

A small operation document may be added as a file-oriented alternative:

~~~yaml
secret_writes:
  - resource: workforce-idp
    fields:
      - config.client_secret
~~~

This would be useful for reviewed CI jobs and avoids long command lines. It is
not required for the first vertical slice if the CLI selectors establish the
underlying intent model.

Write selection belongs to plan generation:

- plan, and direct apply or sync from configuration, may accept it;
- apply or sync of an existing plan must execute only the intents already
  recorded in that artifact;
- write-selection flags used together with --plan must be rejected;
- delete mode must not accept secret-write selection.

## Configuration and source requirements

Write-only secret fields must not accept literal values.

For example, this must fail during loading or validation:

~~~yaml
config:
  client_secret: plaintext-value
~~~

The error must identify the resource and field but must never reproduce the
value. This restriction applies even to test fixtures.

The initial safe source is the existing deferred !env form:

~~~yaml
config:
  client_secret: !env PORTAL_OIDC_CLIENT_SECRET
~~~

File-backed secret support must use a deferred file reference. The current
eager !file behavior is not safe for a write-only secret because the resolved
value can enter a serialized plan. Until a deferred form exists, eager !file
must also be rejected for these fields.

A possible deferred form is:

~~~yaml
config:
  client_secret: !secret-file /run/secrets/portal-oidc
~~~

The final tag name and map syntax can follow existing tag conventions, but its
required behavior is:

- preserve only the source reference in the plan;
- read the file during execution;
- support files mounted by common CI/CD secret mechanisms;
- never serialize the contents into the plan.

Environment and file reference names are metadata, not secret values. Human
output may display only the source kind if exposing names or paths is
undesirable.

Planning a secret write should validate the source declaration, not require
the production value to be available. This allows an unprivileged planning
job to create and review the plan while a separate execution job supplies the
secret. Direct apply or sync performs source preflight and execution in the
same invocation.

Do not automatically convert a rejected literal into an apply-time binding.
Users must deliberately choose a deferred source.

## Plan representation

Extend a planned resource change with explicit secret-write metadata. The
shape below is illustrative:

~~~json
{
  "resource_ref": "workforce-idp",
  "action": "UPDATE",
  "secret_writes": [
    {
      "field": "/config/client_secret",
      "source": {
        "kind": "env",
        "reference": "PORTAL_OIDC_CLIENT_SECRET"
      }
    }
  ]
}
~~~

Required properties:

- The target uses stable declarative field names or JSON-pointer paths.
- The artifact contains the source reference but never the resolved value.
- A secret write is not represented as an old/new comparison.
- The plan can merge a secret write with an ordinary UPDATE for the same
  resource.
- A secret-only intent creates an executable UPDATE even when the normal
  comparison reports no changes.
- A CREATE includes the configured secret once and must not schedule a
  duplicate update.
- Old plan artifacts without secret-write metadata remain readable.

Human plan and diff output should show something equivalent to:

~~~text
UPDATE portal_identity_provider workforce-idp

  config.client_secret: write requested
  current value:        unavailable
  source:               environment reference
~~~

Plan summaries and confirmation prompts should report the number of secret
writes and affected resources without displaying values.

## Planner integration

Keep normal state comparison unchanged. Planning should occur in two phases:

1. Generate the ordinary state-derived plan.
2. Expand explicit secret-write intents.

The intent phase must:

- resolve each selector against the declarative ResourceSet;
- find known write-only fields configured on the selected resource;
- reject an explicitly selected field that is absent;
- never interpret an absent or nil field as a request to clear it;
- verify that the field has a supported write lifecycle;
- attach the write to an existing change or create a secret-only update;
- include non-secret request context required by the resource's API;
- avoid adding another update after a resource CREATE.

A broad --write-secrets request applies to all supported, configured,
write-only fields. It must report create-only or unsupported fields rather
than silently claiming complete coverage. An explicitly selected unsupported
field must fail with an actionable error.

## Execution and security requirements

The plan artifact is a contract for obtaining and writing a secret, not a
secret carrier.

Execution must:

1. preflight every required environment or file source before making any API
   mutation;
2. resolve values only after preflight succeeds;
3. inject the value into a private request payload immediately before the API
   call;
4. avoid replacing source references in the public PlannedChange object with
   plaintext;
5. ensure reporters, errors, debug logs, and HTTP trace logging cannot observe
   resolved secret values;
6. discard resolved values as soon as practical after the request.

Secret-aware HTTP body redaction is required even if the value did not
originate from !env.

Unset sources are errors. Empty-but-present values must not accidentally clear
a secret. They should be rejected by default unless a particular API
explicitly supports clearing and the user requests that operation separately.

Saved plans must never contain literal secret values. Redacting display output
is insufficient if plaintext remains in JSON.

## Resource capability model

Konnect secret fields do not all have the same lifecycle. Resource
implementations need a small reviewed capability declaration containing the
information required to execute a write, such as:

- declarative field path;
- whether it is accepted on create;
- whether an existing value is updateable;
- whether it uses the parent update operation or a dedicated child endpoint;
- request payload construction requirements;
- whether the credential is create-only.

This runtime declaration should live with the relevant resource implementation.
The private sensitive-field catalog can help audit coverage, but kongctl must
not depend at runtime on a large generated audit artifact or blindly trust
OpenAPI writeOnly metadata.

The common user intent may map to different API operations:

- embedded write-only field: update the owning resource;
- config-store or standalone secret: call its create/update/upsert endpoint;
- create-only credential: report that it cannot be updated and requires a new
  credential resource.

Create-only credential rotation is not to be disguised as an UPDATE.

## Initial resource scope

The first implementation should cover the currently declaratively supported
resources whose planners intentionally omit write-only values from comparison:

- Portal OIDC Identity Provider config.client_secret;
- DCR Provider dcr_token, api_key, and initial_client_secret variants;
- AI Gateway Model Provider authentication values;
- AI Gateway Identity Provider config.client_secret;
- AI Gateway Vault secret configuration fields;
- Event Gateway Schema Registry authentication password.

Config-store secret resources should use the same intent and source model once
their declarative resource and endpoint support is available. Fields with
unverified or inconsistent runtime behavior must not be automatically included
until their capability is reviewed.

## Non-goals

- Automatic rotation scheduling or expiration management.
- Determining whether an unreadable remote value actually differs.
- Maintaining a local secret-state or fingerprint lock file.
- Persisting secret hashes in Konnect labels.
- Encrypting plaintext into self-contained plan artifacts.
- General-purpose integrations with external secret managers in the first
  version.
- Automatically rotating or retiring create-only credentials.

A future source-provider interface may add AWS Secrets Manager, HashiCorp
Vault, or other systems without changing the write-intent model.

## Acceptance criteria

- [ ] Known write-only fields reject literal scalar values without echoing them
      in errors.
- [ ] Eager !file values are rejected for those fields unless deferred file
      resolution is implemented.
- [ ] Existing !env references remain unresolved in saved plan artifacts.
- [ ] Environment-backed planning does not require access to the secret value
      when only the source declaration needs validation.
- [ ] Users can request all configured secret writes, all secret writes for one
      resource, or one exact field.
- [ ] Without an explicit intent, existing write-only comparison behavior is
      unchanged.
- [ ] A secret-only request produces an executable planned update.
- [ ] Secret intents merge with ordinary resource updates without duplicate API
      operations.
- [ ] Creates send each configured secret exactly once.
- [ ] Saved plans contain target and source metadata but no resolved values.
- [ ] Applying a saved plan resolves the source at execution time.
- [ ] All secret sources are preflighted before the first API mutation.
- [ ] Plan, diff, confirmation, progress, error, and HTTP trace output never
      disclose resolved values.
- [ ] Explicit create-only or unsupported selections produce actionable errors.
- [ ] Broad selection clearly reports any fields it cannot write.
- [ ] Old plan artifacts remain executable.
- [ ] Unit tests cover selector resolution, literal rejection, serialization,
      redaction, source preflight, intent merging, and missing sources.
- [ ] Adapter tests cover the initial supported resource set.
- [ ] At least one end-to-end scenario verifies planning in one process and
      execution in another with the secret available only during execution.

rspurgeon:
## Summary

- add an explicit `!secret` YAML tag for deferred write-only values, including
  ordered composition such as `Bearer ` plus an environment source
- add `--write-secret` and `--write-secrets` plan-authoring controls for exact
  fields, resources, and broad secret rotations
- make aggregate `--write-secrets` selection best-effort, with skipped or empty
  selections reported on STDERR and retained in saved plan metadata
- record source metadata and write intent in plans without recording resolved
  secret values
- preflight all secret sources and target paths before mutation, inject values
  into private request payloads, and expand HTTP body redaction
- cover Portal IDPs, DCR providers, AI Gateway providers, identity providers
  and vaults, Event Gateway schema registries, and create-only AI Consumer
  Credentials
- normalize selected-secret updates to action-specific SDK payload contracts
- validate the API's one-header limit for AI Gateway Model Provider
  authentication before execution
- add a complete `docs/examples/declarative/secrets` workflow for creates,
  rotations, saved plans, composed values, and aggregate writes
- add an OpenAI LLM quickstart that declaratively creates an AI Gateway,
  provider, model, and data plane certificate, with a local Docker helper that
  never invokes kongctl

## User experience

```yaml
client_secret: !secret {source: !env PORTAL_OIDC_CLIENT_SECRET}
```

Existing resources are written only through an explicit plan intent:

```sh
kongctl plan --mode apply -f konnect.yaml \
  --write-secret "workforce-idp#config.client_secret" \
  --output-file rotation.json
kongctl apply --plan rotation.json --auto-approve
```

Creates automatically send configured secrets once. Existing resources remain
idempotent unless a matching selector is supplied. AI Consumer Credential API
keys are supported on create and rejected as exact update intents because they
are create-only. The aggregate `--write-secrets` flag skips ineligible fields
with warnings while retaining all eligible writes.

## Security model

Planning stores only the resource field, deferred source metadata, and public
composition text. It does not require or serialize resolved secret values.
Execution validates every source and injection target before the first API
mutation, then resolves and injects each value only into a private copy of the
planned change.

The reviewed implementation plan and detailed CLI examples are available in
[issue #1748](https://github.com/Kong/kongctl/issues/1748#issuecomment-5207015065).

## Testing

- `go fix ./...`
- `make format`
- `make build`
- `make lint`
- `make test`
- `make test-integration`
- `bash -n docs/examples/declarative/ai-gateway/openai-llm/data-plane.sh`
- manually verified the secrets workflow and OpenAI LLM quickstart

Closes #1748


internal/declarative/planner/secret_writes.go
```diff
@@ -0,0 +1,494 @@
...
+	if resource.GetType() == resources.ResourceTypeDCRProvider {
+		if providerType, ok := fields[FieldDCRProviderProviderType]; ok {
+			fields[FieldDCRProviderUpdateType] = providerType
+		}
+	}
+	return fields, nil
+}
+
+func mergeSecretContextFields(contextFields, plannedFields map[string]any) map[string]any {
+	merged := make(map[string]any, len(contextFields)+len(plannedFields))
```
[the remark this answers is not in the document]
rspurgeon: Thanks for flagging this. Removed the computed map capacity in commit 9dec313a, so the two potentially attacker-influenced map lengths are no longer added before allocation. The merge behavior is unchanged. Verified with `go test -race ./internal/declarative/planner` and planner-scoped golangci-lint.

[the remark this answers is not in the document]
rspurgeon: Thanks for flagging this. This is the second operand of the same allocation-size expression; commit 9dec313a removes the computed capacity entirely, eliminating both overflow paths while preserving merge behavior. Verified with `go test -race ./internal/declarative/planner` and planner-scoped golangci-lint.

internal/declarative/planner/secret_writes.go
```diff
@@ -0,0 +1,586 @@
...
+			return
+		}
+		removeSecretPathValue(typed[segments[0]], segments[1:])
+	case []any:
+		index, err := strconv.Atoi(segments[0])
+		if err != nil || index < 0 || index >= len(typed) {
+			return
+		}
+		if len(segments) == 1 {
+			typed[index] = nil
```
rspurgeon: **Must fix: this can clear a live secret.**

`removeSecretPathValue` deletes map keys but *nils* array elements. When a declared secret path ends in an array index, the element becomes `null` instead of being removed, and that `null` stays in the outgoing payload.

That case is reachable today, not hypothetical:

- `internal/declarative/secrets/catalog.go:27` declares `/config/client_secret/*` for `ai_gateway_identity_provider`.
- The field really is an array — see this PR's own neighbouring fixture, `internal/declarative/planner/ai_gateway_identity_provider_planner_test.go:44`: `"client_secret": []any{"super-secret"}`.
- `extractAIGatewayIdentityProviderUpdateFields` puts the merged `config` into `Fields` **unscrubbed**. `scrubAIGatewayIdentityProviderSecretFields` is applied only to `ChangedFields` (`ai_gateway_identity_provider_planner.go:301-302`) and to the comparison copies (`:349-350`). So `stripDeclaredSecretPaths` always runs against a live array on an ordinary update.

Reproduction — update change for an AI Gateway OIDC identity provider with **no** `--write-secret` selector:

```
update fields after strip: {"config":{"client_id":"abc","client_secret":[null],"consumer_claims":["sub"]}}
```

and that payload decodes into the SDK's `[]string` as:

```go
json.Unmarshal([]byte(`[null]`), &[]string{})   // => []string{""}
```

Net effect: changing an unrelated field (`display_name`, `consumer_claims`, …) on such a provider sends `client_secret: [""]`. That is the case #1748 explicitly rules out:

> Unset sources are errors. Empty-but-present values must not accidentally clear a secret. They should be rejected by default […]

Nothing catches it downstream. `preflightSecretWrites` only validates target paths for changes that *have* a `SecretWrites` intent, and this change has none — the field was stripped precisely because no write was requested.

**Suggested fix:** remove the element rather than nil it, and drop the parent key when the array empties. Watch the index shift: if two paths on the same array are declared (`.../0` and `.../1`), collect the indices and delete descending, or mark and compact in a single pass, so the second deletion does not target a shifted slot.

**Regression test worth adding:** a table over each catalog path shape asserting the stripped payload contains no residue — no key, no `null`, and no `__SECRET__:` string. `secret_writes_test.go` currently asserts only the *write* path plus one map-shaped `assert.NotContains(config, FieldAPIKey)`, which is why the array shape slipped through.

internal/declarative/loader/secret_sources.go
```diff
@@ -0,0 +1,120 @@
...
+	"github.com/kong/kongctl/internal/declarative/secrets"
+	"github.com/kong/kongctl/internal/declarative/tags"
+)
+
+func (l *Loader) collectSecretSources(actual, placeholder *resources.ResourceSet) error {
+	if actual == nil || placeholder == nil {
+		return nil
+	}
+
+	return visitResourceSetResources(placeholder, func(resource resources.Resource) error {
```
rspurgeon: **Must fix: `!secret` is silently accepted outside reviewed fields here.**

`visitResourceSetResources` is `ResourceSet.ForEachResource`, i.e. the resource registry. `organization.users[]` and `organization.system-accounts[]` are not registered resource types — `env_resolution.go:31` reaches them through a dedicated `resolveOrganizationSelectorEnvPlaceholders`, but this walker has no equivalent. Anything in those collections is never validated, so its `__SECRET__:` placeholder survives loading as a plain string value.

Reproduction against this branch:

```yaml
organization:
  users:
    - ref: u
      email: !secret {source: !env PROBE_EMAIL}
```

```
err=<nil>
org user email=__SECRET__:eyJwYXJ0cyI6W3sic291cmNlIjp7ImtpbmQiOiJlbnYiLCJyZWZlcmVuY2UiOiJQUk9CRV9FTUFJTCJ9fV19
```

That base64 blob is then sent to Konnect as the user's email address.

For contrast, both of these reject correctly, which pins the gap to the un-registered collections rather than to the validation logic:

- `portals[].description: !secret {...}` → `field /description is not a reviewed write-only field and cannot use !secret`
- `_defaults.kongctl.namespace: !secret {...}` → `field /kongctl/namespace is not a reviewed write-only field and cannot use !secret`

This breaks the invariant stated in the approved plan and repeated in `docs/declarative.md`:

> `!secret` is valid only on reviewed write-only fields. Use elsewhere fails.

**Suggested fix, two layers:**

1. Mirror `resolveOrganizationSelectorEnvPlaceholders` here so org users and system accounts are visited and rejected.
2. Add a final sweep after loading that fails if **any** residual `tags.IsSecretPlaceholder` string survives anywhere in the ResourceSet.

Layer 2 is the durable one — it closes every future walker gap instead of just this one, and it is cheap because the traversal already exists. A negative test for a walker-coverage hole would be the first of its kind here; there is currently no test asserting that `!secret` fails outside the registry.

docs/examples/declarative/ai-gateway/config-store-vault.yaml
```diff
@@ -32,4 +32,4 @@ ai_gateways:
             type: basic
             headers:
               - name: Authorization
-                value: "{vault://support-secrets/openai-auth-header}"
+                value: !secret {source: !env OPENAI_VAULT_REFERENCE}
```
rspurgeon: **Must fix (design): Konnect `{vault://...}` references can no longer be expressed.**

This line is the clearest evidence that the literal ban is too broad. Before this PR it read:

```yaml
value: "{vault://support-secrets/openai-auth-header}"
```

A `{vault://...}` string is **not** secret material — it is a public pointer, and the whole point of the AI Gateway config-store / vault feature is that *Konnect* resolves it so the value never touches kongctl or the plan artifact. Rewriting it to `!secret {source: !env OPENAI_VAULT_REFERENCE}` inverts the intent: a non-secret reference now has to be smuggled through an environment variable at execution time, and it disappears from the reviewable plan. The example also no longer demonstrates the feature it exists to document.

The practical consequence is broader than the example. `internal/declarative/loader/secret_sources.go:56-59` rejects the literal during **loading**, so any user following Kong's recommended vault workflow on a catalog field has a config that no longer works with `plan`, `diff`, `apply`, `sync` or `delete`.

**Suggested fix:** allow vault-reference literals (`{vault://...}`, plus any sibling syntax Konnect accepts) on catalog fields, and restore this example to its original value. It is a reference, not key material, and treating it as a secret makes the user's posture strictly worse.

Two related items with the same root cause, worth resolving together:

- **`internal/declarative/secrets/catalog.go:23-34` globs wider than what was reviewed.** The approved plan enumerates concrete paths — `config.auth.headers[].value`, `config.auth.client_secret`, `config.auth.secret_access_key`, `config.auth.aws.secret_access_key`, `config.auth.service_account_json`. The catalog instead ships `/config/**/...`, and `**` matches any depth, so every field named `key`, `token` or `secret_id` anywhere below `config` becomes a mandatory `!secret`. An AI Gateway vault's `config.key`, which names *which* key to fetch, is the obvious false positive. A glob that captures unreviewed paths undercuts the catalog's own premise that it is "reviewed manually rather than inferred from field names." The matcher already supports `*` for array indices, which is all the reviewed set needs.
- **Literal rejection has no escape hatch and no deprecation window,** while bare `!env` — a strictly safer form — gets a full cycle (`applySecretWriteIntents` warns and normalizes it). That asymmetry is backwards, and it is the concern raised in the issue thread: *"When deck encounters an issue similar to this (hashed basic auth) it writes a warning rather than erroring out entirely. I think that applying files containing secrets as an update will be a common workflow."* The `--write-secrets` behaviour was softened in response; literal rejection was not. This PR is already carrying one known integration fixture failure caused by it. One release of value-safe warnings before rejecting would match the courtesy bare `!env` receives.

rspurgeon (commented):
## Security review: must-fix findings

Reviewed the secret-write model end to end (loader → catalog → planner → executor → HTTP redaction). **The core security model holds where it matters most:** plans carry source references and never resolved values, `preflightSecretWrites` runs before the first mutation, and `executeChange` clones the change before injection so reporters and the plan object never observe plaintext. Those are the right boundaries in the right places.

Three findings break that model at its edges and are flagged as must-fix. Two were reproduced by running throwaway tests against this branch (`3b2c224`); each inline comment includes the reproduction.

| # | Finding | File |
|---|---|---|
| 1 | Stripping an array-backed secret leaves `null`, which reaches Konnect as `""` — can clear a live secret | `internal/declarative/planner/secret_writes.go:429` |
| 2 | `!secret` silently accepted on `organization.users` / `system-accounts`, shipped as a literal placeholder | `internal/declarative/loader/secret_sources.go:17` |
| 3 | Konnect `{vault://...}` references can no longer be expressed | `docs/examples/declarative/ai-gateway/config-store-vault.yaml:35` |

Findings 1 and 2 are correctness defects with a security consequence. Finding 3 is a design decision that breaks existing configuration on upgrade — the evidence is a conversion in this PR's own diff.

Quality gates on this branch are clean, so none of these are caught today:

- `go test ./internal/...` — pass
- `golangci-lint run --new-from-rev=origin/main` — 0 issues

Non-blocking findings (plan-output fidelity for secret-only updates, nondeterministic change IDs from map iteration in `applySecretWriteIntents`, redaction/catalog drift, adapter and `resolveSecretResourceID` test gaps) are held back so this review stays focused on the blockers. Happy to file them separately.

internal/declarative/planner/secret_writes.go
rspurgeon: Fixed in 417c2e9d. Declared paths are now ordered so array indexes are removed highest-to-lowest, and array entries are compacted instead of being replaced with nil. When the last secret entry is removed, the parent array field is removed as well. The regression test covers both Fields and ChangedFields and verifies the serialized change has no client_secret array containing null and no secret placeholder residue.

internal/declarative/loader/secret_sources.go
rspurgeon: Fixed in 417c2e9d. Successfully collected secret-tag placeholders are cleared from the disposable placeholder resource set, then a final walk of the entire resource set rejects any residual placeholder. This makes the check durable for resources omitted from the registry traversal. Regression coverage includes both organization user email and organization system-account name selectors.

docs/examples/declarative/ai-gateway/config-store-vault.yaml
rspurgeon: Fixed in 417c2e9d. The AI Gateway catalog now contains only the reviewed concrete provider and vault paths; recursive matching and the false-positive vault config key entry were removed. Recognized public Konnect vault references now pass through as ordinary, plan-visible configuration, and the Config Store/Vault example has been restored to a literal reference. The missed model-matrix fixture was migrated to the secret tag as well. I kept the hard error for actual literal secret material: accepting it even temporarily would allow the value to enter a saved plan, which is the security boundary this feature establishes.

PR #1949, opened 2026-08-22
fix(executor): treat api document title as optional on create

### LINKED ISSUE #1947: Bug: Declarative API document scaffold produces plan rejected by apply
## Summary

A declarative API document generated from kongctl scaffold api.documents passes loading and planning, but its plan is rejected by kongctl apply because the executor requires title. The explain schema and scaffold both represent title as optional.

Observed with kongctl 1.13.0, commit a33d30b2.

## Reproduction

1. Run kongctl explain api.documents -o json.
2. Run kongctl scaffold api.documents.
3. Use the active scaffold fields for a nested API document: content, slug, and ref. Leave the scaffolded title comment disabled.
4. Generate a plan with kongctl plan -f config.yaml --mode apply --output-file plan.json.
5. Run kongctl apply --plan plan.json --auto-approve -o json.

## Actual behavior

Planning succeeds and emits an api_document CREATE change without title. Apply then fails before making the request:

    incompatible plan change 2 ("3:c:api_document:<ref>") for CREATE api_document: title is required; regenerate the plan

The generated explain schema requires content and ref, but not title. The generated scaffold has:

    - content: !file ./content.txt
      # title: value
      slug: my-resource
      ref: my-resource

The current resource validation also checks content but not title, while APIDocumentAdapter.RequiredFields and MapCreateFields require both title and content.

## Expected behavior

The declarative contract should be consistent across explain, scaffold, validation, planning, and apply. Based on the SDK model and current resource reference, title appears to be optional, so apply should accept the plan without it. If title is intended to be required, explain should list it as required, scaffold should activate it, and validation/planning should reject the input before producing an executable plan.

## Impact

A user can bootstrap a seemingly valid declaration using kongctl scaffold and successfully generate and diff a plan, only to have apply reject that same plan as incompatible.

yzxcj797:
Fixes #1947.

## Root cause

Every layer of the declarative pipeline treats an API document's `title` as **optional** — the SDK's `CreateAPIDocumentRequest.Title` is a `*string`, the generated explain schema does not list it as required, `scaffold api.documents` emits it commented out, and `APIDocumentResource.Validate()` only requires `content`. Only the executor disagreed: `MapCreateFields` returned `title is required` and `RequiredFields()` listed both — so a plan built verbatim from the scaffold passed loading, validation and planning, then failed `apply` with `incompatible plan change ... title is required; regenerate the plan`.

## Fix

`APIDocumentAdapter`:

- `MapCreateFields` maps `title` only when present (the same optional pattern the update path already uses); `content` remains required;
- `RequiredFields()` returns only `content`, matching the explain schema and resource validation.

## Testing

New `api_document_adapter_test.go`:

- **CreateWithoutTitle** — the scaffold-shaped field set (content/slug/ref, no title) maps cleanly, title stays nil;
- **CreateWithTitle** — title flows through when present;
- **CreateStillRequiresContent** — content-only requirement pinned (`RequiredFields() == ["content"]`).

`go test ./internal/declarative/executor/ -run TestAPIDocumentAdapter` passes; differential with the adapter reverted fails on the title-required error. The full `./internal/...` suite: 76 packages ok; `internal/cmd/output/columns` and `internal/cmd/root/verbs/dump` fail identically on `main` in this Windows environment (pre-existing, unrelated).

### DIFF
--- a/internal/declarative/executor/api_document_adapter.go
+++ b/internal/declarative/executor/api_document_adapter.go
@@ -26,12 +26,6 @@ func (a *APIDocumentAdapter) MapCreateFields(
 	create *kkComps.CreateAPIDocumentRequest,
 ) error {
 	// Required fields
-	title, ok := fields[planner.FieldTitle].(string)
-	if !ok {
-		return fmt.Errorf("title is required")
-	}
-	create.Title = &title
-
 	content, ok := fields[planner.FieldContent].(string)
 	if !ok {
 		return fmt.Errorf("content is required")
@@ -39,6 +33,13 @@ func (a *APIDocumentAdapter) MapCreateFields(
 	create.Content = content
 
 	// Optional fields
+	// The API (and the explain schema and scaffold) treat title as
+	// optional: CreateAPIDocumentRequest.Title is a pointer. Requiring it
+	// here rejected plans built straight from the scaffold output (#1947).
+	if title, ok := fields[planner.FieldTitle].(string); ok {
+		create.Title = &title
+	}
+
 	if slug, ok := fields[planner.FieldSlug].(string); ok {
 		create.Slug = &slug
 	}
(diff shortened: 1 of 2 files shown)


rspurgeon:
@yzxcj797 thank you for the contribution. We ask athat all contributors to please sign the communit license agreement. After that I can review and proceed w/ this.  Thanks again

rspurgeon:
Replaced by #1972

PR #1991, opened 2026-08-26
fix(declarative): prevent AI Gateway perpetual drift

### LINKED ISSUE #1987: apply: ai_gateway_vault and ai_gateway_model_provider report perpetual drift with declarative config-store secrets (prerelease-1-14-0)
## Summary

`kongctl apply` against an AI Gateway config with declarative `config_stores`/`vaults`/`secrets` (new in `prerelease-1-14-0`) never converges. `ai_gateway_vault` and `ai_gateway_model_provider` are reported as `<configuration changes detected>` and re-applied on **every single run**, even with zero changes to the config file or environment between runs.

## Version

`prerelease-1-14-0` (installed via `curl -fsSL https://get.konghq.com/kongctl | sh -s -- --version prerelease-1-14-0`)

## Repro

1. Apply this minimal config (throwaway `ai_gateway` with a declarative config-store secret + vault + model provider):

```yaml
_defaults:
  kongctl:
    namespace: drift-repro

ai_gateways:
  - ref: drift-repro-gateway
    name: drift-repro-gateway
    display_name: "kongctl vault/provider drift repro"
    proxy_urls:
      - host: drift-repro.example.com
        port: 443
        protocol: https
    config_stores:
      - ref: repro-config-store
        name: repro-config-store
        display_name: Repro-Config-Store
        secrets:
          - ref: repro-secret
            key: repro-secret-key
            value: !secret {source: !env REPRO_SECRET_VALUE}
    vaults:
      - ref: repro-vault
        name: repro-vault
        type: konnect
        config:
          config_store_id: !ref repro-config-store#id
    model_providers:
      - ref: repro-openai-provider
        name: repro-openai-provider
        type: openai
        display_name: "OpenAI Provider (repro)"
        config:
          auth:
            type: basic
            headers:
              - name: Authorization
                value: "{vault://repro-vault/repro-secret-key}"
```

```bash
export REPRO_SECRET_VALUE="repro-value"
kongctl apply -f config.yaml --pat "$PAT" --base-url https://us.api.konghq.tech --auto-approve
```

2. Re-run the identical command again (no edits to the file or `REPRO_SECRET_VALUE`):

```bash
kongctl apply -f config.yaml --pat "$PAT" --base-url https://us.api.konghq.tech --auto-approve
```

## Expected

Second (and every subsequent) `apply` reports `0 changes` — the resources are already in the declared state.

## Actual

Every re-run — verified 4 consecutive times with zero changes in between — reports the same 2 resources as changed and re-executes their update:

```
ai_gateway_vault (1 resources):
  ~ test-vault
    <configuration changes detected>

ai_gateway_model_provider (1 resources):
  ~ openai-provider
    <configuration changes detected>
    depends on: ai_gateway_vault:test-vault

SUMMARY
----------------------------------------------------------------------
  Total changes: 2
  Resources to update: 2
```

This never converges — it's not a one-time settle-then-stable transition, it's a permanent diff on the same two resource types.

`ai_gateway`, `ai_gateway_config_store`, and `ai_gateway_config_store_secret` do **not** exhibit this — only `ai_gateway_vault` and `ai_gateway_model_provider` (the one whose `auth.headers[].value` contains the `{vault://...}` reference) drift.

## Why this matters

We use `kongctl apply` inside an automated test harness that reapplies declarative config repeatedly (retries, relaunches, a 3x-identical-run determinism gate). A resource type that reports spurious changes on every apply breaks that gate and makes real regressions indistinguishable from this noise — the same failure shape as the existing `_deck` gateway-entity non-idempotency issue, just now reproduced on the `ai_gateway_vault`/`ai_gateway_model_provider` resource types specifically.

## Suspicion

The vault's `config.config_store_id` is set via `!ref repro-config-store#id` — the plan output includes a warning `Reference config.config_store_id=__REF__:test-config-store#id will be resolved during execution`, so my guess is the diff engine is comparing the unresolved `__REF__:...` placeholder (or some other pre-resolution representation) against the live resource's resolved UUID on every plan, and the provider's `{vault://...}` header value diffs similarly because it depends on the vault. Un-verified — flagging as a starting point, not a root-cause claim.

rspurgeon:
## Summary

- treat AI Gateway write-only values omitted from API responses as unknown during comparison
- normalize declarative config-store references to resolved Konnect IDs before comparing vault configuration
- preserve comparison of vault references when both desired and current payloads expose them
- add reapply convergence coverage and catalog-wide secret-write planning tests

## Testing

- go fix ./...
- make format
- make build
- make test
- golangci-lint run ./internal/declarative/planner/...
- make test-e2e-scenarios SCENARIO=ai-gateway/config-store

Closes #1987

### DIFF
--- a/internal/declarative/planner/ai_gateway_auth_strategy_planner.go
+++ b/internal/declarative/planner/ai_gateway_auth_strategy_planner.go
@@ -297,9 +297,10 @@ func shouldUpdateAIGatewayAuthStrategy(
 	}
 
 	if desired.Config != nil && aiGatewayAuthStrategyConfigChanged(current.Config, desired.Config) {
+		currentConfig, desiredConfig := comparableAIGatewayAuthStrategyConfigs(current.Config, desired.Config)
 		changedFields[FieldConfig] = FieldChange{
-			Old: scrubAIGatewayAuthStrategySecretFields(normalizeAuthStrategyConfigForCompare(current.Config)),
-			New: scrubAIGatewayAuthStrategySecretFields(normalizeAuthStrategyConfigForCompare(desired.Config)),
+			Old: currentConfig,
+			New: desiredConfig,
 		}
 	}
 
@@ -343,17 +344,28 @@ func extractAIGatewayAuthStrategyFields(provider resources.AIGatewayAuthStrategy
 }
 
 func aiGatewayAuthStrategyConfigChanged(current, desired map[string]any) bool {
+	currentComparable, desiredComparable := comparableAIGatewayAuthStrategyConfigs(current, desired)
+	return !reflect.DeepEqual(currentComparable, desiredComparable)
+}
+
+func comparableAIGatewayAuthStrategyConfigs(current, desired map[string]any) (map[string]any, map[string]any) {
+	return comparableAIGatewayWriteOnlyPayloads(
+		current,
+		desired,
+		normalizeAIGatewayAuthStrategyConfigsForComparison,
+		scrubAIGatewayAuthStrategySecretFields,
+		isAIGatewayAuthStrategySecretField,
+	)
+}
+
+func normalizeAIGatewayAuthStrategyConfigsForComparison(
+	current map[string]any,
+	desired map[string]any,
+) (map[string]any, map[string]any) {
 	currentComparable := normalizeAuthStrategyConfigForCompare(current)
 	desiredComparable := normalizeAuthStrategyConfigForCompare(desired)
 	projectAIGatewayAuthStrategyConfigForComparison(currentComparable, desiredComparable)
-	currentComparable = scrubAIGatewayAuthStrategySecretFields(currentComparable).(map[string]any)
-	desiredComparable = scrubAIGatewayAuthStrategySecretFields(desiredComparable).(map[string]any)
-	pruneUnpairedAIGatewayWriteOnlyReferences(
-		currentComparable,
-		desiredComparable,
-		isAIGatewayAuthStrategySecretField,
-	)
-	return !reflect.DeepEqual(currentComparable, desiredComparable)
+	return currentComparable, desiredComparable
 }
 
 func projectAIGatewayAuthStrategyConfigForComparison(currentCompare map[string]any, desiredCompare map[string]any) {
--- a/internal/declarative/planner/ai_gateway_auth_strategy_planner.go
+++ b/internal/declarative/planner/ai_gateway_auth_strategy_planner.go
@@ -348,6 +348,11 @@ func aiGatewayAuthStrategyConfigChanged(current, desired map[string]any) bool {
 	projectAIGatewayAuthStrategyConfigForComparison(currentComparable, desiredComparable)
 	currentComparable = scrubAIGatewayAuthStrategySecretFields(currentComparable).(map[string]any)
 	desiredComparable = scrubAIGatewayAuthStrategySecretFields(desiredComparable).(map[string]any)
+	pruneUnpairedAIGatewayWriteOnlyReferences(
+		currentComparable,
+		desiredComparable,
+		isAIGatewayAuthStrategySecretField,
+	)
 	return !reflect.DeepEqual(currentComparable, desiredComparable)
 }
 
--- a/internal/declarative/planner/ai_gateway_payload_compare.go
+++ b/internal/declarative/planner/ai_gateway_payload_compare.go
@@ -103,6 +103,35 @@ func scrubAIGatewayUpstreamWriteOnlyFields(value any) any {
 	}
 }
 
+type aiGatewayPayloadPairNormalizer func(map[string]any, map[string]any) (map[string]any, map[string]any)
+
+func scrubbedAIGatewayWriteOnlyPayloads(
+	current map[string]any,
+	desired map[string]any,
+	normalize aiGatewayPayloadPairNormalizer,
+	scrub func(any) any,
+) (map[string]any, map[string]any) {
+	currentComparable, desiredComparable := normalize(current, desired)
+	return scrub(currentComparable).(map[string]any), scrub(desiredComparable).(map[string]any)
+}
+
+func comparableAIGatewayWriteOnlyPayloads(
+	current map[string]any,
+	desired map[string]any,
+	normalize aiGatewayPayloadPairNormalizer,
+	scrub func(any) any,
+	isWriteOnlyField func(string) bool,
+) (map[string]any, map[string]any) {
+	currentComparable, desiredComparable := scrubbedAIGatewayWriteOnlyPayloads(
+		current,
+		desired,
+		normalize,
+		scrub,
+	)
+	pruneUnpairedAIGatewayWriteOnlyReferences(currentComparable, desiredComparable, isWriteOnlyField)
+	return currentComparable, desiredComparable
+}
+
 func pruneUnpairedAIGatewayWriteOnlyReferences(
 	current any,
 	desired any,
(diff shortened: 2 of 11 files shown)


internal/declarative/planner/ai_gateway_provider_planner.go
```diff
@@ -284,9 +284,10 @@ func shouldUpdateAIGatewayProvider(
 	}
 
 	if desired.Config != nil && aiGatewayProviderConfigChanged(current.Config, desired.Config) {
+		currentConfig, desiredConfig := comparableAIGatewayProviderConfigs(current.Config, desired.Config)
 		changedFields[FieldConfig] = FieldChange{
-			Old: scrubAIGatewayProviderSecretFields(normalizeProviderConfigForCompare(current.Config)),
-			New: scrubAIGatewayProviderSecretFields(normalizeProviderConfigForCompare(desired.Config)),
+			Old: currentConfig,
+			New: desiredConfig,
```
rspurgeon, line 295: **Bug:** the displayed `changedFields[FieldConfig]` now reuses the *pruned* comparable maps, so it can silently hide a genuinely new write-only field from the plan/diff output whenever it's added alongside another real config change.

Example: `current.Config = {"auth":{"type":"basic"}}`, `desired.Config = {"auth":{"type":"oauth2","client_secret":"{vault://.../x}"}}`. `aiGatewayProviderConfigChanged` correctly reports `true` (type differs), but `comparableAIGatewayProviderConfigs` prunes the unpaired `client_secret` before it's shown here, so the plan/diff never reveals that a vault-secret reference was added. The actual applied payload (`extractAIGatewayProviderFields`) is unaffected — only the user-visible diff loses information.

Suggest computing the displayed Old/New from a scrub-only (unpruned) pipeline, separate from the pruned pipeline used for the changed/unchanged decision.

internal/declarative/planner/ai_gateway_auth_strategy_planner.go
```diff
@@ -348,6 +348,11 @@ func aiGatewayAuthStrategyConfigChanged(current, desired map[string]any) bool {
 	projectAIGatewayAuthStrategyConfigForComparison(currentComparable, desiredComparable)
 	currentComparable = scrubAIGatewayAuthStrategySecretFields(currentComparable).(map[string]any)
 	desiredComparable = scrubAIGatewayAuthStrategySecretFields(desiredComparable).(map[string]any)
+	pruneUnpairedAIGatewayWriteOnlyReferences(
```
rspurgeon: **Bug:** this pruning is applied to the change-detection comparison inside `aiGatewayAuthStrategyConfigChanged`, but the *displayed* `changedFields[FieldConfig].Old/New` built in `shouldUpdateAIGatewayAuthStrategy` (this file, around the `scrubAIGatewayAuthStrategySecretFields(normalizeAuthStrategyConfigForCompare(...))` calls) never got the equivalent fix — it skips both `projectAIGatewayAuthStrategyConfigForComparison` and this new `pruneUnpairedAIGatewayWriteOnlyReferences` call.

Concretely: if `client_secret` is set via a stable vault reference and only `issuer` changes, the change is correctly attributed to `issuer`, but the rendered plan/diff will still show `client_secret` appearing/disappearing between Old and New — reproducing the exact perpetual-drift-looking noise this PR is meant to eliminate, just in the rendered diff instead of the plan decision. Unlike the sibling provider planner (see other comment on this review), which now shares one `comparableAIGatewayProviderConfigs` helper for both paths, this file's two paths have diverged.

internal/declarative/planner/ai_gateway_payload_compare.go
```diff
@@ -103,6 +103,47 @@ func scrubAIGatewayUpstreamWriteOnlyFields(value any) any {
 	}
 }
 
+func pruneUnpairedAIGatewayWriteOnlyReferences(
```
rspurgeon, line 135: Minor: this is now a fourth hand-written "prune keys missing from peer" tree-walker alongside `pruneAIGatewayDefaultsMissingFromPeer` and `pruneEmptyContainersMissingFromPeer` above, with a different mutation contract (mutates both `current` and `desired` in a single call, vs. the existing convention of calling the other helpers once per direction). It's also only wired into 3 of the AI Gateway resource types that have write-only secret fields. Not blocking, but worth a follow-up to consolidate, especially since the same class of bug (perpetual drift from an unpaired write-only field) likely exists for any resource type this isn't wired into yet.

internal/declarative/planner/ai_gateway_vault_planner.go
```diff
@@ -236,6 +240,40 @@ func shouldUpdateAIGatewayVault(
 	return true, desiredPayload, changedFields, nil
 }
 
+func comparableAIGatewayVaultPayloads(current, desired map[string]any) (map[string]any, map[string]any) {
```
rspurgeon, line 243: Minor: `comparableAIGatewayVaultPayloads` and `comparableAIGatewayProviderConfigs` (ai_gateway_provider_planner.go) are near-identical copy-pasted pipelines (normalize → scrub current → scrub desired → prune-unpaired), differing only in which scrub function and write-only predicate get passed in. Worth factoring into one shared helper parameterized on those two functions so future pipeline changes (e.g. fixing the auth-strategy display-path gap noted above) only need to happen in one place.

rspurgeon (commented):
Reviewed the write-only secret drift fix. The core idea (prune fields that are unpaired because the API omits write-only values) is sound and the vault config-store-ref normalization + reapply convergence tests are a nice addition. Found two correctness issues where the new pruning logic and the plan's *displayed* diff have drifted out of sync with each other (inline comments below), plus two lower-priority duplication notes.

internal/declarative/planner/ai_gateway_payload_compare.go
rspurgeon: Good point. Commit 2da74248 consolidates the normalize/scrub/prune pipeline used by provider, auth strategy, and vault. The catalog-wide planner test also verifies every update-capable ai_gateway secret capability: config-store secrets use their dedicated write-only resource path, consumer credentials are create-only, and agent/MCP upstream credentials are unconditionally scrubbed. The lower-level missing-peer walkers remain separate because their pruning predicates and directional semantics differ.

internal/declarative/planner/ai_gateway_provider_planner.go
rspurgeon: Thanks — fixed in commit 2da74248. Provider change detection still uses the pruned comparison, while the displayed config diff now uses a scrubbed-but-unpruned projection so a vault reference added alongside an observable config change remains visible. Added a regression test for the basic-to-OAuth example and verified the planner suite plus the model-provider E2E scenario.

internal/declarative/planner/ai_gateway_auth_strategy_planner.go
rspurgeon: Thanks — fixed in commit 2da74248. Auth-strategy change detection and displayed Old/New values now share the same normalized, projected, scrubbed, and pruned comparison output. Added a regression test proving an omitted client_secret vault reference does not appear as diff noise when issuer changes.

internal/declarative/planner/ai_gateway_vault_planner.go
rspurgeon: Thanks — addressed in commit 2da74248. Provider, auth strategy, and vault now delegate to one shared comparableAIGatewayWriteOnlyPayloads pipeline parameterized by the pair normalizer, scrubber, and write-only predicate. Provider display projection separately opts out of unpaired pruning as required by the other finding.

PR #2116, opened 2026-09-08
chore(ci): update the KSAI workflows for ksai/v3.42.1

bartsmykla:
## Motivation

The KSAI workflows this repository calls have changed shape since v1.15.0, v1.2.0, v1.3.0. Moving the pin alone is not enough: an input a caller passes that the workflow no longer declares, or a permission it does not grant, is refused by GitHub before any job starts and reports nothing a reader can act on.

This carries the change and the pins together, which is why it is one pull request rather than a version bump.

`Kong/ksai` is internal and this repository is public, so the calls resolve to `Kong/ksai-public`, the generated snapshot published for exactly this case.

## Implementation

- The job is replaced by calls to the published workflows rather than running the action directly, so the runner, the timeout and the concurrency lane are the called workflow's from now on. Any `concurrency:` block this file had is removed for that reason, and the workflow is renamed to `KSAI`
- All three calls are pinned by commit digest at `v3.42.1`
- `anthropic_base_url` is passed as `${{ vars.KSAI_ANTHROPIC_BASE_URL }}`. The snapshot redacts Kong-internal hostnames, so its copy of that input carries no default and is refused empty rather than falling back to `api.anthropic.com`. A variable rather than a literal, so the hostname stays out of this public repository
- `app_login: kong-api-ops` is passed to the `ksai-hold` call, naming the App `client_id` already names. `v3.41.0` added that input, and the snapshot carries no default for it, which holds a pull request carrying a commit any App published - Renovate's and Dependabot's included. Naming the App means only KSAI's own commits hold one
- `authorization_github_token`, `bot_login`, `ci_evidence_github_token`, `comment_body`, `continuation_attempt`, `continuation_prev_remaining`, `continuation_stall`, `kong_org_github_token`, `pr_number`, `source_org_github_token` are not carried over. The published workflow computes those values itself and passes them to the action, so setting them on the caller is no longer possible or needed
- `allowed_commands` is not carried over: it was replaced by `disabled_commands`, which names the commands this repository turns off rather than the ones it answers, so an allowlist cannot be carried across without inverting it. It was empty here, so nothing was narrowed and nothing is lost
- `ksai_ref` is not carried over: it names which `Kong/ksai` commit a local-path caller checks out, and a caller of the published workflow is pinned by its own `uses:` line, so the adopted shape has nothing to point it at
- `job_timeout_minutes` and `require_plan_approval` are passed unquoted. `ksai.yml` declares them `type: number` and `type: boolean`, and the quoted strings this file carried over made every run a `startup_failure` - which produces no annotation and never reaches `gh pr checks`
- `.github/workflows/ksai-assign.yaml` is deleted. It dispatched `ksai.yaml` on the same team `review_requested` event the `request` job above now answers through `ksai-request.yml`, so both dispatched; the reviewer shares one `cancel-in-progress` concurrency group per pull request, so the second could cancel a review already running. It also had no bot-sender exclusion, so the App's own team request re-triggered it
- There is no `ksai-actions` preset this repository can extend, so none was added to your Renovate config. Until there is one, Renovate cannot tell one release target from another, and this will not be the last pull request of its kind
- Everything else this repository had set is kept: its runner, its inputs, its secrets and its comments. Only the lines above change

> [!WARNING]
> **The `if:` on `gate` and `review` and `implement` and `clear-request` is replaced by the published workflow's own.**
>
> If yours narrowed what starts a run, this widens it, and every pull request the new one admits spends your Anthropic budget. Yours was:
>
> ```yaml
> if: >- (github.event_name == 'workflow_dispatch' && inputs.pr != '') || (github.event_name == 'issue_comment' && github.event.comment.user.type != 'Bot' && contains(github.event.comment.body, '/ksai'))
> if: ${{ needs.gate.outputs.authorized == 'true' && needs.gate.outputs.review == 'true' }}
> if: >- !cancelled() && ((github.event_name == 'workflow_dispatch' && inputs.issue_number != '') || (needs.gate.result == 'success' && needs.gate.outputs.authorized == 'true' && needs.gate.outputs.implement == 'true' && contains(fromJSON('["OWNER", "MEMBER", "COLLABORATOR"]'), github.event.comment.author_association)))
> if: >- always() && needs.gate.result == 'success' && (needs.review.result == 'success' || (github.event_name == 'workflow_dispatch' && (needs.gate.outputs.authorized != 'true' || needs.gate.outputs.review != 'true')))
> ```
>
> `require_label`, `skip_label`, `review_drafts` and `review_bots` are inputs on that call and say the same thing without an expression to maintain.

> [!IMPORTANT]
> **Set `KSAI_ANTHROPIC_BASE_URL` before this merges, or every run fails before it starts.**
>
> ```bash
> gh variable set KSAI_ANTHROPIC_BASE_URL --body '<the origin every model call goes to>' --repo Kong/kongctl
> ```
>
> The Kong organization already sets it, so nothing is needed here unless this repository overrides it.

> [!NOTE]
> **`write_access_commands` defaults to `all` from `v3.41.0`, where it was empty before.**
>
> This file does not set it, so it takes the new default: every command needs write access on this repository and no longer needs the commenter to own a path in `CODEOWNERS`. Holding `fix` and `review` to ownership kept contributors working across team lines out of both, and anyone holding write can already push whatever those commands would write. Set `write_access_commands: ""` on the `ksai` call to keep the stricter bar.

> [!NOTE]
> **`client_id` names this repository's own App, so that App must grant four read permissions.**
>
> `contents: read`, `issues: read`, `members: read` and `pull requests: read`. GitHub refuses the whole token request when one of them is missing and never says which, so a run dies at `run / gate` naming none of them. Newly added permissions also stay pending on an existing installation until an organization owner accepts them.

`/ksai` works as it does today, with the wider command bar noted above.


.github/workflows/ksai.yaml
```diff
@@ -1,715 +1,138 @@
...
+      clear_request: true
+      pr: ${{ inputs.pr }}
+      actor: ${{ inputs.actor }}
+      issue_number: ${{ inputs.issue_number }}
+      work_ref: ${{ inputs.work_ref }}
+      attempt: ${{ inputs.attempt }}
+      stall: ${{ inputs.stall }}
+      prev_remaining: ${{ inputs.prev_remaining }}
+      job_timeout_minutes: "35"
+      require_plan_approval: "true"
```
rspurgeon: @bartsmykla **[P1] Pass the declared number and boolean input types.**

The pinned ksai.yml declares job_timeout_minutes as a number and require_plan_approval as a boolean, but this caller passes quoted strings ("35" and "true"). Reusable workflow input values must match the declared types, so these values prevent the workflow from validating. Change them to unquoted 35 and true.

Contract: https://github.com/Kong/ksai-public/blob/4c9c72ddd664c76f6ca9706d67122a2515a6cd32/.github/workflows/ksai.yml#L213-L218 and https://github.com/Kong/ksai-public/blob/4c9c72ddd664c76f6ca9706d67122a2515a6cd32/.github/workflows/ksai.yml#L257-L262. GitHub documents the matching-type requirement at https://docs.github.com/en/actions/how-tos/reuse-automations/reuse-workflows.

.github/workflows/ksai.yaml
```diff
@@ -1,715 +1,138 @@
...
-      ((github.event_name == 'workflow_dispatch' && inputs.issue_number != '') ||
-       (needs.gate.result == 'success' &&
-        needs.gate.outputs.authorized == 'true' &&
-        needs.gate.outputs.implement == 'true' &&
-        contains(fromJSON('["OWNER", "MEMBER", "COLLABORATOR"]'), github.event.comment.author_association)))
+      (github.event_name == 'pull_request' &&
+      github.event.action == 'review_requested' &&
+      github.event.requested_team.slug == (vars.KSAI_REVIEW_TEAM || 'ksai') &&
+      github.event.pull_request.head.repo.full_name == github.repository &&
+      github.event.sender.type != 'Bot') ||
```
rspurgeon, line 104: @bartsmykla **[P2] Retire the existing dispatcher when adding this replacement.**

.github/workflows/ksai-assign.yaml remains enabled and dispatches ksai.yaml for the same same-repository team review_requested event. A human sidebar request therefore triggers both that workflow and this new request job. The reusable request workflow also creates team requests with an App token for automatic reviews; those events trigger the old dispatcher, which has no bot-sender exclusion, in addition to the reusable workflow dispatching directly.

These paths start duplicate reviews, and the called reviewer uses the same PR concurrency group with cancel-in-progress: true, so a duplicate can cancel a review already running. Remove or disable the old dispatcher as part of this migration so each request dispatches once.

rspurgeon (changes requested):
@bartsmykla Two migration issues need attention: the reusable workflow receives incorrectly typed inputs, and the retained ksai-assign.yaml duplicates the new review dispatcher. Details are inline.

Validation: local unit tests passed with CGO_ENABLED=0 go test -count=1 ./...; actionlint passed with ShellCheck disabled because the local shim has no configured version. E2E scenarios are not applicable to this workflow-only change. The current commit also has KSAI startup_failure runs that are absent from gh pr checks, including https://github.com/Kong/kongctl/actions/runs/34226495550.

PR #2128, opened 2026-09-09
feat(mesh): core command surface for Kong Mesh control planes

justindavies:
# kongctl + Kong Mesh: user story and walkthrough

Every command below was run against a live Konnect-hosted global control plane
with two federated zone CPs (`zone1`, `zone2`) on Kubernetes. Output is copied
from those runs, not composed. Test resources were created and deleted as part
of the walkthrough; both clusters and the global CP are back to their starting
state.

## Who this is for

A platform engineer whose Kong Mesh global control plane is hosted in Konnect.
The global CP is Kong's infrastructure — they have no `kubectl` access to it and
no shell on it. A CLI is the only way they can read or write global state.

That framing matters for reviewing this PR: for a Konnect user, these commands
aren't a convenience over `kubectl`. They are the only option.

### 1. Find the control plane

After `kongctl login`:

```
$ kongctl get mesh control-planes
NAME  ID     API LINE
v3    d820…  v3
```

### 2. Ask the control plane what it serves

kongctl carries no built-in list of Kong Mesh resource types. It asks the
control plane, and renders what comes back:

```
$ kongctl get mesh resource-types
NAME                    ALIAS   SCOPE   KIND      WRITABLE
dataplanes              dp      Mesh    resource  no
meshaccesslogs          mal     Mesh    policy    yes
meshes                  m       Global  resource  yes
...
37 types
```

This is the main design decision in the PR. When a Kong Mesh release adds a
policy type, it appears here with no kongctl change. `SCOPE`, `KIND` and
`WRITABLE` are all the control plane's answers — `WRITABLE: no` on `dataplanes`
is reported by the CP, not inferred by kongctl.

### 3. Write once, land in every zone

```
$ kongctl apply mesh -f mesh.yaml
TYPE  NAME                 RESULT
Mesh  kongctl-walkthrough  created

$ kongctl apply mesh -f timeout-policy.yaml
TYPE         NAME                  MESH                 RESULT
MeshTimeout  walkthrough-timeouts  kongctl-walkthrough  created
```

The policy was stored at global with `kuma.io/origin: global`, and both zone
clusters received it through KDS:

```
# zone1
$ kubectl get meshtimeout -A
NAMESPACE          NAME                                    TARGETREF KIND
kong-mesh-system   walkthrough-timeouts-vxb4d24vwc8w9vc4   Mesh

# zone2 — identical
$ kubectl get meshtimeout -A
NAMESPACE          NAME                                    TARGETREF KIND
kong-mesh-system   walkthrough-timeouts-vxb4d24vwc8w9vc4   Mesh
```

One apply reached both zones. The alternative is applying the same manifest
against every zone cluster by hand.

Re-applying the same file reports `updated` rather than failing, matching the
upsert semantics of `kumactl apply` and `kubectl apply`:

```
$ kongctl apply mesh -f mesh.yaml
TYPE  NAME                 RESULT
Mesh  kongctl-walkthrough  updated
```

### 4. Read it back

```
$ kongctl get mesh meshtimeouts --mesh kongctl-walkthrough
MESH                 NAME                  AGE
kongctl-walkthrough  walkthrough-timeouts  11s

$ kongctl get mesh meshtimeouts walkthrough-timeouts --mesh kongctl-walkthrough -o yaml
kri: kri_mt_kongctl-walkthrough___walkthrough-timeouts_
labels:
    kuma.io/display-name: walkthrough-timeouts
    kuma.io/mesh: kongctl-walkthrough
    kuma.io/origin: global
mesh: kongctl-walkthrough
name: walkthrough-timeouts
spec:
    targetRef:
        kind: Mesh
    to:
        - default:
            connectionTimeout: 5s
            idleTimeout: 1h0m0s
          targetRef:
            kind: Mesh
type: MeshTimeout
```

### 5. Export

```
$ kongctl dump mesh --export-profile federation > export.yaml
# 23 resources
```

Profiles: `all`, `federation`, `federation-with-policies`, `no-dataplanes`.
The federation profiles are for seeding a *different* global control plane.

### 6. Tokens

```
$ kongctl create mesh zone-token --zone zone1 --valid-for 1h
eyJhbGciOiJSUzI1NiIsImtpZCI6IjEiLCJ0eXAiOiJKV1QifQ...

$ kongctl create mesh dataplane-token --mesh default --name test-dp --valid-for 1h
eyJhbGciOiJSUzI1NiIsImtpZCI6IjEiLCJ0eXAiOiJKV1QifQ...
```

The zone token's payload decodes to `{"Zone":"zone1","Scope":["cp"]}` — the
`cp` scope is what the KDS auth validator requires of a zone CP connecting to
global (`pkg/kds/auth/tokens/zone_validator_adapter.go`).

### 7. Delete

```
$ kongctl delete mesh meshtimeouts walkthrough-timeouts --mesh kongctl-walkthrough
TYPE         NAME                  MESH                 RESULT
MeshTimeout  walkthrough-timeouts  kongctl-walkthrough  deleted

$ kongctl delete mesh meshes kongctl-walkthrough
TYPE  NAME                 RESULT
Mesh  kongctl-walkthrough  deleted
```

Deletion propagates to the zones over KDS. Propagation is not instant — the
zone clusters still showed the resources immediately after the delete returned,
and were clear within a minute. Worth knowing when scripting against this.

## Environment

- Global CP: Konnect-hosted, API line v3
- Zones: `zone1`, `zone2`, both Kubernetes, both federated and connected
- Control plane build: `kong/kuma-cp:0.0.0-preview.v075f9c95c`
- Verified against `kumahq/kuma@origin/master` and `Kong/kong-mesh@origin/master`


internal/cmd/root/products/konnect/mesh/client.go
```diff
@@ -0,0 +1,201 @@
...
+}
+
+// resolveBaseURL determines which control plane a command addresses.
+//
+// meshcommon resolves an explicit URL or an identifier from configuration
+// alone. A name needs a Konnect call to become an identifier, which cannot
+// live there, so it is resolved here and the identifier written back to
+// configuration — leaving the composition itself in one place.
+func resolveBaseURL(helper cmd.Helper, cfg config.Hook) (string, error) {
+	baseURL, err := meshcommon.ResolveControlPlaneAPIURL(cfg)
```
rspurgeon, line 455: **High: honor explicit control-plane selection over configured defaults.**

A configured `konnect.mesh.control-plane.id` makes this resolver return before considering `--control-plane-name`. I reproduced `--control-plane-name new-name` sending requests to configured `old-id`. This resolver also serves writes and deletes, so this can operate on a different control plane than the operator selected.

Make explicit selection override configured alternative selectors, or reject conflicting selections with an actionable error. Include command-level regression tests for ID/name/URL combinations across flags, environment variables, and configuration.

internal/cmd/root/products/konnect/mesh/mesh.go
```diff
@@ -0,0 +1,168 @@
...
+	if addParentFlags != nil {
+		addParentFlags(verb, baseCmd)
+	}
+	meshcommon.AddControlPlaneFlags(baseCmd.PersistentFlags())
+	if verb == verbs.Create {
+		baseCmd.Flags().StringSliceP(FilenameFlagName, "f", nil,
+			"Files, directories, URLs, or - for stdin, holding the mesh resources to apply. Repeatable.")
+	}
+	if verb == verbs.Dump {
+		baseCmd.Flags().String(ProfileFlagName, ProfileFederation,
```
rspurgeon: **High: preserve the global configuration/authentication `--profile` flag.**

This local flag shadows kongctl's global profile option. `kongctl dump mesh --profile prod` fails with `invalid profile "prod"` because it interprets the configuration profile as an export selection.

Rename this option, for example to `--export-profile`, and retain the global profile behavior. Test selecting a non-default configuration profile together with each export selection.

internal/cmd/root/products/konnect/mesh/getResources.go
```diff
@@ -0,0 +1,153 @@
...
+	if len(args) > 1 {
+		name = args[1]
+	}
+
+	path, err := requestPath(cfg, descriptor, name)
+	if err != nil {
+		return &cmd.ConfigurationError{Err: err}
+	}
+
+	body, err := fetch(helper, path)
```
rspurgeon: **Medium: paginate resource listings completely.**

This performs only one collection request. In a local mock check, a response with one item, `total: 2`, and a next link caused a successful exit without fetching the remaining item. Text output also hides the pagination metadata, making the result appear complete.

Aggregate all pages consistently with existing get operations, while retaining single-resource reads. Inspection collections have the same single-fetch pattern and need equivalent handling. Cover multi-page responses and termination conditions with HTTP integration tests.

internal/cmd/root/products/konnect/mesh/createResources.go
```diff
@@ -0,0 +1,357 @@
...
+		return false, fmt.Errorf(
+			"%s is read only on this control plane and cannot be created or updated", descriptor.Singular())
+	}
+
+	mesh := resource.Mesh
+	if !descriptor.IsMeshScoped() {
+		mesh = ""
+	}
+
+	status, err := sendForStatus(helper, http.MethodPut, descriptor.ItemPath(mesh, resource.Name), resource.Body)
```
rspurgeon: **Medium: make replacement intent explicit instead of silently updating under `create`.**

Every document is sent with PUT, so an existing resource can be replaced and reported as `updated` without the operator explicitly requesting replacement. The comment explaining why declarative `apply` was avoided does not resolve the surprising behavior of `create`.

Establish explicit replacement semantics, such as an opt-in overwrite operation or a clearly named update operation, and document them in user-facing help. Test existing-resource handling and partial failures, including an E2E create/get/delete lifecycle.

internal/cmd/root/products/konnect/mesh/client.go
```diff
@@ -0,0 +1,201 @@
...
+	if err != nil {
+		return nil, 0, err
+	}
+
+	baseURL, err := resolveBaseURL(helper, cfg)
+	if err != nil {
+		return nil, 0, err
+	}
+
+	tokenSource, err := konnectcommon.GetAccessTokenSource(cfg, logger)
```
rspurgeon: **Medium: separate self-managed authentication from Konnect authentication.**

The command advertises `--control-plane-url` for self-managed control planes, but this path unconditionally resolves Konnect credentials. I verified that an unauthenticated local control plane receives no request: the CLI fails first with `resolve Konnect access token`.

Support the appropriate self-managed authentication modes, keeping credentials scoped to their intended destination, or narrow the advertised capability until that support exists. Add authentication-path coverage for both hosted and self-managed targets.

internal/cmd/root/verbs/get/get.go
```diff
@@ -208,6 +208,12 @@ Setting this value overrides tokens obtained from the login command.
 	}
 	cmd.AddCommand(regionsCmd)
 
+	meshCmd, err := NewDirectMeshCmd()
+	if err != nil {
+		return nil, err
+	}
+	cmd.AddCommand(meshCmd)
```
rspurgeon, line 217: **Medium: register the explicit Konnect command paths as well.**

`kongctl get mesh` is registered here, but `kongctl get konnect mesh` fails with `unknown command "mesh"`. The new constructors' comments promise both direct and explicit forms, matching the existing product convention.

Register the supported Mesh operations consistently in both trees and test the actual root command paths, not only the Mesh constructors or helpers.

internal/cmd/root/products/konnect/mesh/client.go
```diff
@@ -0,0 +1,201 @@
...
+		headers map[string]string
+	)
+	if body != nil {
+		payload = bytes.NewReader(body)
+		headers = map[string]string{"Content-Type": "application/json"}
+	}
+
+	result, err := apiutil.RequestWithTokenSource(
+		ctx,
+		httpclient.NewLoggingHTTPClient(logger),
```
rspurgeon: **Reuse the existing configuration-aware HTTP client construction structures.**

`NewLoggingHTTPClient(logger)` constructs a client with a fixed default timeout and default transport options. Mesh requests therefore bypass the configured HTTP behavior used by existing Konnect operations. The same pattern appears in control-plane listing and resource input downloads.

Reuse the existing timeout/transport resolvers (`ResolveHTTPTimeout`, `ResolveHTTPTransportOptions`), `httpclient.ClientConfig`, and logging wrapper; reuse or extract a shared builder where necessary rather than maintaining another construction policy. Preserve credential separation for input downloads. Add tests proving configured timeout and transport settings reach the client.

internal/cmd/root/products/konnect/mesh/createTokens.go
```diff
@@ -0,0 +1,281 @@
...
+	if err != nil {
+		return err
+	}
+
+	validFor, err := requireValidFor(cmdObj)
+	if err != nil {
+		return err
+	}
+
+	name, err := cmdObj.Flags().GetString(tokenNameFlagName)
```
rspurgeon, line 187: **Complete and document the configuration contract for Mesh options.**

The common control-plane/mesh flags are correctly bound through `config.Hook`, but token-specific options are read directly from Cobra here. Export selection and inspection type have the same pattern, so they have no corresponding configuration-file/environment-variable support.

For options intended to support persistent defaults, define documented configuration paths, use the shared flag-binding pattern, and read effective values through configuration. Explicitly document invocation-only inputs. Where configuration can satisfy a required value, validate the resolved value rather than relying only on `MarkFlagRequired`. Add file/environment/flag precedence tests.

internal/cmd/root/products/konnect/mesh/dumpResources.go
```diff
@@ -0,0 +1,292 @@
...
+			}
+		case ProfileNoDataplanes:
+			if descriptor.Name == "Dataplane" || descriptor.Name == "DataplaneInsight" {
+				continue
+			}
+		}
+		selected = append(selected, descriptor)
+	}
+
+	sort.SliceStable(selected, func(i, j int) bool {
```
rspurgeon: **Follow the repository's Go 1.26 modernization standards.**

Use `slices.SortStableFunc` here instead of `sort.SliceStable`, and `slices.Sort` for remaining necessary string sorts. Run `go fix ./...` followed by the repository's formatting and quality gates when making the revisions.

Also simplify `displayTags` in `printers.go`: each value is a singleton slice, the keys are already sorted, and the subsequent value/output sorting adds unnecessary work and complexity.

internal/cmd/root/products/konnect/mesh/getResources_test.go
```diff
@@ -0,0 +1,75 @@
...
+// stubConfig returns a config hook answering only the given paths, so a test
+// states exactly the configuration it depends on.
+func stubConfig(values map[string]string, flags map[string]bool) *configtest.MockConfigHook {
+	return &configtest.MockConfigHook{
+		GetStringMock: func(key string) string { return values[key] },
+		GetBoolMock:   func(key string) bool { return flags[key] },
+	}
+}
+
+func TestRequestPathScoping(t *testing.T) {
```
rspurgeon, line 20: **Add complete command, HTTP integration, and Mesh E2E scenario coverage.**

These helper tests are useful, but the added suite does not exercise complete Mesh command wiring or HTTP flows. The reproduced profile collision, missing explicit product path, selector precedence bug, and first-page-only listing all escape the current tests.

Please add command tests for routing/flag validation, configuration precedence tests, and HTTP integration tests for pagination, authentication, configured clients, and write failures. Add isolated Mesh E2E scenarios covering discovery/read, create/get/delete, export/reapply, and supported inspection/token flows, with cleanup, and run them through the trusted E2E workflow. The existing generic smoke/version scenario does not validate this feature.

rspurgeon (changes requested):
## Review: changes requested

Reviewed commit `144018bdc2958bb2edcf0681577594de155e1e8f`.

The discovery-driven resource design is useful, and reuse of existing output infrastructure is appropriate. However, target selection, command semantics, configuration integration, and feature-level test coverage need revisions before merge. The inline comments provide the specific locations and requested changes.

### Correctness and UX

- A configured control-plane ID overrides an explicit control-plane name. Because the resolver serves writes and deletes, this can select the wrong target.
- The export `--profile` option shadows the global configuration/authentication profile.
- Resource listings and inspection collections do not paginate completely.
- `create` can replace an existing resource without explicit replacement intent.
- Advertised self-managed support unconditionally requires Konnect authentication.
- The explicit `get konnect mesh` command path is missing despite the promised direct/explicit convention.

### Configuration and HTTP construction

The common ID/name/URL, mesh, and all-meshes options use `config.Hook`; PAT/base URL use existing configuration paths. Local mock checks confirmed mesh selection honors flag > environment > configuration file > default, and the all-meshes environment setting works. The target-selection defect concerns precedence between alternative selectors, not a general failure to load configuration.

Token-specific options, inspection type, and export selection are read directly from Cobra. Define and document configuration paths for persistent defaults, explicitly identify invocation-only inputs, and validate resolved required values where configuration is supported.

Reuse the existing configuration-aware HTTP construction structures instead of constructing default clients: shared timeout/transport resolution, `httpclient.ClientConfig`, and logging/authentication composition as appropriate. Reuse or extract a shared builder rather than introducing another construction policy. Keep input-download credentials separate from control-plane credentials.

### Testing required

Add complete command-path tests, configuration precedence tests, HTTP integration coverage, and dedicated Mesh E2E scenarios. Include pagination, hosted/self-managed authentication, configured timeout/transport settings, existing-resource replacement behavior, and partial write failures. E2E coverage should include discovery/read, create/get/delete, export/reapply, and supported inspection/token flows, using isolated resources and cleanup.

### Modernization, dependencies, and scope

- Follow the repository's Go 1.26 modernization standards, including `slices.SortStableFunc`/`slices.Sort`, and run `go fix ./...` plus the required formatting and quality gates after revisions.
- Simplify the unnecessary singleton slices and repeated sorting in tag rendering.
- No new dependency/version conflicts were found: `go.mod` and `go.sum` are unchanged.
- Consider separating export/migration and inspection into follow-up PRs. The currently empty description does not establish why the full 4,688-line surface belongs in foundational support.
- Use the required `type(category): title` PR title and provide scope, rationale, and validation details.

### Validation performed

- Passed: CGO-disabled build, full unit suite, integration suite, generic smoke/version E2E scenario, and local CLI/mock checks.
- Race tests were not run because they require CGO; the repository's CGO-disabled requirement prevents the race-enabled Makefile target from running as written.
- Local lint reported five findings in unchanged files, using golangci-lint 2.13.2 rather than the repository-specified 2.10.1. None pointed to the Mesh additions.
- No dedicated Mesh E2E scenarios exist in this PR; live Mesh behavior remains unverified.
- At review time, GitHub CI tests passed, while trusted E2E and CLA remained pending.

The review left the repository unchanged.

internal/cmd/root/products/konnect/mesh/createResources.go
```diff
@@ -0,0 +1,405 @@
...
+	}
+
+	cfg, err := helper.GetConfig()
+	if err != nil {
+		return nil, err
+	}
+
+	// The configured timeout and transport apply here too; what stays separate
+	// is the credential, which apiutil.Request does not attach.
+	client, err := newHTTPClient(cfg, logger)
```
rspurgeon, line 228: **High: keep control-plane TLS credentials and trust settings out of remote input downloads.**

This client builder installs the self-managed control plane's client certificate, private CA, and `tls-skip-verify` setting. Omitting the bearer token from `apiutil.Request` does not isolate TLS authentication.

I reproduced this with a local HTTPS input server: `apply mesh -f https://<input-server>/input.yaml` presented the configured control-plane client certificate to the input server. Its self-signed server certificate was also accepted through the control-plane skip-verification setting.

Reuse common timeout/transport construction, but separate destination-specific TLS identity and trust policy. A remote input download must not inherit the control plane's client certificate or skip-verification option. Add an HTTP/TLS integration test asserting that separation.

internal/cmd/root/products/konnect/mesh/client.go
```diff
@@ -0,0 +1,533 @@
...
+	ctx := helper.GetContext()
+	if ctx == nil {
+		ctx = context.Background()
+	}
+
+	// A self managed control plane authenticates its own callers, so Konnect
+	// credentials are neither required nor sent. Resolving them regardless
+	// meant an unauthenticated local control plane never received a request:
+	// the command failed first on the missing Konnect token.
+	selfManaged := isSelfManaged(cfg)
```
rspurgeon, line 342: **High: derive authentication and TLS settings from the resolved target.**

The URL resolver now honors an explicit hosted ID/name over a saved self-managed URL, but `isSelfManaged(cfg)` still independently checks whether that URL exists in configuration.

I reproduced a profile with a saved self-managed URL/token and a Konnect PAT, then ran `get mesh meshes --control-plane-id new-id`. Requests reached `/v3/mesh/control-planes/new-id/...` with the self-managed token instead of the Konnect credential. The same classification also controls TLS settings.

Resolve the target once, including whether it is hosted or self-managed, and consistently derive its URL, credentials, and TLS policy from that result. Cover explicit hosted selectors overriding a configured self-managed URL with complete request-level tests.

internal/cmd/root/products/konnect/konnect.go
```diff
@@ -201,6 +203,39 @@ func preRunE(c *cobra.Command, args []string) error {
...
+	cmd *cobra.Command,
+	verb verbs.VerbValue,
+	addParentFlags func(verbs.VerbValue, *cobra.Command),
+	parentPreRun func(*cobra.Command, []string) error,
+) error {
+	if !slices.Contains(meshVerbs, verb) {
+		return nil
+	}
+
+	meshCmd, err := mesh.NewMeshCmd(verb, addParentFlags, parentPreRun)
```
rspurgeon, line 231: **Medium: bind Mesh flags on the explicit Konnect command path.**

This passes the general Konnect pre-run handler, which does not call `meshcommon.BindFlags`. Registering the command makes help work, but the actual Mesh flags do not reach configuration.

Reproduced:
`get konnect mesh meshes --control-plane-url <local-server>`
fails with `no Kong Mesh control plane selected`, while the direct `get mesh` form works. With existing configuration, ignored flags can instead leave the command using configured defaults.

Compose the Konnect and Mesh bindings for this path, including token subcommands. Test actual requests through both direct and explicit command trees, verifying that selectors and command-specific settings take effect.

internal/cmd/root/products/konnect/mesh/mesh.go
```diff
@@ -0,0 +1,163 @@
...
+
+// appliesResources reports whether a verb sends resource documents from -f.
+//
+// Both apply and create do. `apply` is the primary form because it is what
+// kumactl calls this and what the behaviour actually is: the control plane
+// addresses a resource by type and name and a write creates or replaces it,
+// which kumactl implements as an upsert and reports as created or updated.
+// `create` is kept because it was the name this shipped under first.
+func appliesResources(verb verbs.VerbValue) bool {
+	return verb == verbs.Apply || verb == verbs.Create
```
rspurgeon, line 62: **Medium: adding apply does not resolve replacement under create.**

`apply mesh` is a better name for upsert behavior, but `create mesh -f` still reaches the identical write implementation. I reproduced a PUT to an existing resource followed by a successful exit and `"result": "updated"`.

Remove the redundant resource-create alias or make replacement intent explicit. The comment that this is retained because it "shipped under" create needs clarification for an unmerged feature; compatibility should not preserve surprising behavior without an established requirement. Token issuance can remain under create.

Also reconcile the older comment in `runCreateResources` that says the operation uses create rather than apply.

internal/cmd/root/mesh_command_paths_test.go
```diff
@@ -0,0 +1,47 @@
...
+		{"create", "mesh", "--help"},
+		{"create", "konnect", "mesh", "--help"},
+		{"delete", "mesh", "--help"},
+	}
+
+	for _, args := range paths {
+		path := strings.Join(args[:len(args)-1], " ")
+
+		t.Run(path, func(t *testing.T) {
+			result := executeRootForTest(t, args...)
```
rspurgeon, line 29: **Add request-level command tests and dedicated Mesh E2E scenarios.**

The additional tests are useful, but these command-path tests execute only `--help`, which bypasses the bindings and requests that currently fail on the explicit Konnect path. Selector tests exercise flag detection rather than the combined URL/authentication decision. `TestMeshOptionPrecedence` also claims environment coverage but contains no environment-variable case.

Add complete command/HTTP tests for direct and explicit paths, selector/authentication combinations, file/environment/flag precedence, pagination, TLS credential isolation, and write behavior. Add isolated Mesh discovery/read, apply/get/delete, and token E2E scenarios with cleanup, and run them through trusted E2E. There are still no Mesh scenarios in this PR; passing smoke/version does not validate the feature.

internal/cmd/root/verbs/apply/mesh.go
```diff
@@ -0,0 +1,104 @@
...
+	}
+
+	meshCmd.Example = fmt.Sprintf(`  # Apply mesh resources from a file
+  %[1]s apply mesh -f policy.yaml --control-plane-id <id>
+
+  # Apply every resource in a directory
+  %[1]s apply mesh -f ./policies --control-plane-id <id>
+
+  # Apply from stdin, which is how an export is reapplied
+  %[1]s dump mesh --control-plane-id <id> | %[1]s apply mesh -f - --control-plane-id <id>`, meta.CLIName)
```
rspurgeon, line 75: **Update the example after splitting export out of this PR.**

This example invokes `dump mesh`, which is no longer available in this change. Use an available stdin example, and introduce export/reapply examples with the follow-up export feature. Please also reconcile stale apply/create comments so the help and implementation describe one consistent command contract.

rspurgeon (changes requested):
## Follow-up review: substantial progress, changes still requested

Reviewed updated commit `76761c57891b86606344472d333398427d4f0791`.

Moving export and inspect out is appropriate. Follow-up features depending on this foundation is expected, but the remaining correctness issues and missing feature-level coverage should be addressed before merging. The inline comments provide reproductions and specific requested changes.

### Progress against the previous review

| Area | Assessment |
| --- | --- |
| Export/inspect scope and export profile collision | Removed from this PR |
| Resource pagination | Fixed; a local mock confirmed both pages are returned |
| HTTP client construction | Now uses existing configuration resolvers and client structures; destination-specific TLS isolation needs correction |
| Self-managed authentication | Unauthenticated, bearer-token, and TLS support added; target/authentication selection is still inconsistent |
| Token configuration | Lifetime/scope now support configuration; direct-path environment lifetime confirmed |
| Explicit control-plane selection | Original name-over-configured-ID case fixed; hosted selection over a configured self-managed URL still uses the wrong authentication mode |
| Explicit Konnect commands | Registered, but Mesh flags are not bound on that path |
| Create replacement semantics | Unresolved: adding apply retains the same upsert behavior under create |
| Modernization/simplification | The reported sorting and tag-rendering issues are addressed |
| Test coverage | Improved helper/help tests, but still no Mesh HTTP-flow or E2E scenarios |
| PR metadata | Title fixed; description remains empty |

### Remaining findings

1. **High: remote input downloads inherit control-plane TLS identity and trust settings.** A local HTTPS input server received the configured control-plane client certificate, and the control-plane skip-verification setting accepted its self-signed certificate. Keep destination-specific TLS settings separate while reusing general client construction.
2. **High: explicit hosted selection still uses self-managed authentication.** A saved self-managed URL causes authentication/TLS classification to disagree with the newly resolved hosted URL. The local reproduction sent the saved self-managed token to the hosted control-plane endpoint instead of the Konnect credential.
3. **Medium: explicit Konnect commands ignore Mesh flags.** The new registration passes a pre-run handler that does not bind Mesh configuration. The explicit URL command fails with no control plane selected; existing configuration can instead mask the ignored flags.
4. **Medium: create still replaces existing resources.** The identical PUT implementation remains under both apply and create. Establish explicit replacement semantics or remove the redundant resource-create alias.

### Coverage and cleanup

Please add complete command/HTTP tests for the above cases and dedicated Mesh E2E scenarios covering discovery/read, apply/get/delete, and token operations, using isolated resources and cleanup.

The added help tests cannot detect missing runtime bindings. Selector helper tests do not cover the combined target/authentication decision, and the configuration precedence test does not actually exercise environment variables. Those gaps explain why these regressions survive the passing suite.

Remove the stale `dump mesh` example from apply help, reconcile contradictory apply/create comments, and fill in the PR description with scope, rationale, and validation. No dependency changes or conflicts were found.

### Validation performed

- Passed locally: CGO-disabled build, full unit suite, integration suite, lint with zero findings, generic smoke/version E2E, and targeted CLI/mock checks.
- Race tests were not run with CGO disabled.
- No live Mesh E2E scenario coverage exists in this PR.
- At evaluation time, GitHub CI tests passed; trusted E2E and CLA remained pending.

The rewritten branch was evaluated in a separate checkout, preserving the original checkout. No source changes were made during the review.

Issue #2053, opened 2026-09-02
Task(e2e): reduce scenario validation latency and queueing


rspurgeon:
## Summary

Reduce the queue-to-result latency of the `.com` Konnect E2E scenario suite
without weakening scenario coverage or allowing two jobs to mutate the same
organization concurrently.

Implement this as a staged rollout with independently measurable and
revertible pull requests. Do not combine every stage into one large change.

## Isolation contract

Treat the Konnect organization as the unit of isolation:

- Each matrix job receives the PAT for exactly one organization through its
  GitHub environment.
- All `kongctl` and reset calls in that job use that PAT.
- A scenario must not introduce or switch to a second Konnect credential.
- Manually provisioned fixtures such as organization users are read-only
  identities. Scenario-created assignments and roles remain scoped to the
  organization represented by the PAT.
- Scenarios are self-contained and must not depend on another scenario's
  output. `resetOrg` establishes a clean organization when required.

Under this contract, jobs targeting different organizations may overlap even
when different workflow revisions assign scenarios differently. Jobs targeting
the same organization must remain serialized.

## Current behavior and evidence

The workflow currently serializes the complete `.com` E2E workflow through a
top-level concurrency group in `.github/workflows/e2e.yaml`. It also has the
more precise per-organization job group:

```yaml
concurrency:
  group: konnect-e2e-${{ matrix.org_name }}
  queue: max
  cancel-in-progress: false
```

The workflow-level lock was introduced by #1137 after GitHub's former
one-pending-job concurrency behavior caused partial runs. The per-org groups
now use `queue: max`, so they can retain all pending jobs without serializing
unrelated organizations or the build/harness jobs.

The suite currently contains 165 scenarios:

- 153 scenarios contain at least one `resetOrg` command.
- There are 166 `resetOrg` commands in total.
- A reset scans at least 14 resource families, plus configured user-assignment
  state where applicable.
- 12 scenarios are pinned to `kongctl-acceptance`.

The sharding implementation in `test/e2e/scenarios_shard.go` removes pinned
scenarios, modulo-distributes the remaining scenarios by sorted position, and
then appends pinned scenarios to their assigned environment. With the current
suite this gives `kongctl-acceptance` approximately 31 ordinary scenarios plus
12 pinned scenarios, while the other orgs receive approximately 30-31 each.

Count is not a sufficient proxy for duration. In successful run
https://github.com/Kong/kongctl/actions/runs/33655125875, scenario execution
[...]

PR #2229, opened 2026-09-18
task(e2e): parallelize reset inventory reads

rspurgeon:
## Summary

Related to #2053; follows #2226. Reduce live E2E reset overhead by
inventorying at most three resource categories concurrently, while retaining
sequential, dependency-ordered deletion. No CLI behavior, replay eligibility,
shard allocation, or workflow organization lock changes.

## Behavior and safeguards

- Reset configured E2E user assignments first, as before.
- Use three inventory workers, with a separate HTTP session per category.
  Pagination, backoff, request deadlines, and transport recovery remain
  sequential within each category.
- Wait for all inventory reads before deleting anything from those categories.
  Preserve global/regional routing, region skips, filters, identifiers, and
  portal custom-domain cleanup immediately before portal deletion.
- Never delete from partial/failed inventory. Continue other categories and
  report the first error in reset order, independent of completion order.
- Re-list after deletion conflicts rather than retrying stale inventory.
  Preserve existing handling of resources that disappear after inventory.
- Honor the total reset deadline during reads and deletions, cancel active
  requests, and prevent queued categories from issuing requests after cancel.
- Keep actual reset elapsed time separate from overlapping per-category work.
  Retain request/retry counts and durations without changing artifact schemas.
- Add `make test-e2e-harness` to PR CI and `make test-all`: race-enabled,
  credential-free harness tests using local HTTP servers, not live Konnect.

## Validation

- Go modernizations checked; unrelated automatic rewrites excluded.
- `make format`, `make build`, `make lint`, `make test`, and
  `make test-integration` passed.
- `make test-e2e-harness` passed; reset tests additionally passed with
  `-race -count=10`.
- `make test-e2e-metrics` passed (100 tests); default cassette check passed.
- Reset utility builds with `CGO_ENABLED=0` and the E2E tag.
- Installer tests and applicable pre-commit checks, including secret detection
  and YAML parsing, passed. Legacy tab/YAML formatting hooks were skipped to
  preserve existing Go, Makefile, and workflow formatting.
- Additional E2E-tagged lint reports no new findings against the base commit;
  full tagged lint has existing harness findings outside this change. Standard
  repository lint passes with zero issues.

The added synthetic benchmark uses 13 empty categories with a fixed 10 ms
server delay. Five iterations measured ~135 ms/reset with sequential category
scheduling versus ~53 ms/reset with three workers (~61% lower). This isolates
scheduling overhead and is **not** a forecast of real Konnect/CI savings.

```sh
go test -tags=e2e ./test/e2e/harness -run '^$' \
  -bench BenchmarkResetResourcesInventoryLatency -benchtime=5x
```

## Rollout and measurement

Live validation runs through the normal PR E2E workflow with existing org
concurrency locks; no unlocked local reset or new organization is required.
Replay scenarios still skip reset, and main still runs all scenarios live.

After merge, compare reset wall time per reset and longest-shard execution
time for comparable pre/post-change PR and main runs, identifying cohorts by
commit ancestry. Do not sum concurrent list durations as elapsed reset time.
Check list-call counts and errors for increased retries/rate limiting. Keep
historically expected scenario timeouts and workflow retries separate from
claims about this optimization's correctness or measured savings.

The conservative starting cap is three reads per organization (up to 15 across
five live shards). Increasing it or parallelizing writes is out of scope.


rspurgeon:
Fixed the replay failure from run 35383200024 in this PR.

The document PATCH exactly matched recorded exchange 132, but could arrive
before API inventory exchange 127 completed. The matcher treated the unmet
dependency as a request mismatch. A 100 ms delay of that inventory response
reproduced the exact CI failure at interaction 129, ending at 131/268.

The proxy now reserves an exact-match exchange inside the current approved
parallel phase and waits up to five seconds for its dependencies. The wait
releases the condition lock so other handlers can satisfy those dependencies.
Matching, recorded dependency rules, phase boundaries, and exactly-once
consumption remain strict. Timeout diagnostics identify missing interaction
numbers without exposing request data. Failure/shutdown wakes pending handlers.

Validation:

- Regression test failed with the original interaction-129 error before the
  fix and passes afterward, consuming the complete unchanged cassette.
- The original 100 ms delayed-lookup probe now passes all 268 exchanges.
- All ten enabled scenarios passed local replay. These local runs were not
  network-namespace isolated; the normal PR job provides isolated validation.
- Seven focused concurrency tests passed 20 consecutive iterations (140 tests).
- All 107 E2E tooling tests passed, along with build, standard lint, Go unit,
  integration, race-enabled harness, installer, and applicable pre-commit
  checks, including secret detection.

No cassette re-recording, scenario changes, replay-policy membership changes,
or weakened assertions. The original reset-concurrency change is unchanged.

